PR #5, opened 2021-09-09
Configure an ansible playbook

### LINKED ISSUE #4: Create Ansible Playbook
As an administrator,
I want an Ansible Playbook for Poseidon,
in order to have a simple, self-documented deployment process

mpass99:
Closes #4

### DIFF
--- a/deploy/ansible/playbook.yml
+++ b/deploy/ansible/playbook.yml
@@ -1,0 +1,82 @@
+---
+
+# Run this playbook via e.g.: ansible-playbook -i inventory-staging.yaml playbook.yml
+- name: Build and Deploy Poseidon
+  connection: ssh
+  gather_facts: false
+  hosts: all
+  tasks:
+    # Build Poseidon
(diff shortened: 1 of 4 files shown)


deploy/ansible/playbook.yml
```diff
@@ -0,0 +1,82 @@
...
+  connection: ssh
+  gather_facts: false
+  hosts: all
+  tasks:
+    # Build Poseidon
+    - name: Build Poseidon
+      local_action:
+        module: make
+        chdir: "{{ playbook_dir }}/../.."
+        target: build
```
mpass99, line 14: Currently we compile Poseidon on the local computer / controller. Although this has the advantage that we do not need all the files and dependencies required for compiling on the Poseidon server, this can also break down under certain architecture combinations...

MrSerth: I am fine with compiling Poseidon locally and pushing the binary to the server as part of the deployment process.

PR #6, opened 2021-09-29
Change authorization header key

mpass99:
as the use of headers starting with X- has been deprecated in RFC6648.

### DIFF
--- a/configuration.example.yaml
+++ b/configuration.example.yaml
@@ -4,7 +4,7 @@ server:
   address: 127.0.0.1
   # Port on which the webserver listens
   port: 7200
-  # If set, this token is required in the Poseidon-Token header for each route except /health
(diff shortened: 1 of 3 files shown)


configuration.example.yaml
```diff
@@ -4,7 +4,7 @@ server:
   address: 127.0.0.1
   # Port on which the webserver listens
   port: 7200
-  # If set, this token is required in the X-Poseidon-Token header for each route except /health
+  # If set, this token is required in the Poseidon-Token header for each route except /health
```
MrSerth, line 7:
```suggestion
  # If set, this token is required in the `Poseidon-Token` header for each route except /health
```

MrSerth (approved):
Thanks a lot (especially for the fixes in the documentation)!

MrSerth (approved):

PR #11, opened 2021-10-08
#9 Implement routes to list, get and delete execution environments.

### LINKED ISSUE #9: Lost routes of the swagger API
Implement
- List route: Return all execution environments in Poseidon's memory.
  - query parameter to additionally synchronise with Nomad
- Get route: Return the details of an execution environment according to the information in Poseidon's memory.
  - query parameter to additionally synchronise with Nomad
- Delete route

mpass99:
Closes #9 

A refactoring was required to introduce the ExecutionEnvironment interface.

- So far no abstraction to the Nomad job exists.
  - Whenever you wanted to get information, you had to search the TaskGroups and tasks.
  - Complicated code to get simple information such as the prewarming pool size
  - Now there is a simple interface so that access to environment details is now self-documenting and clear.

- Even the runner management interface contained a Nomad job as a parameter.
  - This would prevent Poseidon from using other technologies besides Nomad in the future
  - Now a technology can be realised simply by implementing the interface

### DIFF
--- a/internal/api/environments_test.go
+++ b/internal/api/environments_test.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"github.com/gorilla/mux"
 	"github.com/openHPI/poseidon/internal/environment"
+	"github.com/openHPI/poseidon/internal/nomad"
+	"github.com/openHPI/poseidon/internal/runner"
 	"github.com/openHPI/poseidon/pkg/dto"
 	"github.com/openHPI/poseidon/tests"
 	"github.com/stretchr/testify/mock"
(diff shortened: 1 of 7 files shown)


mpass99:
ToDo: Add more tests

mpass99:
Thanks for the truly good comments :ok_hand:

internal/api/environments_test.go
```diff
@@ -32,6 +33,140 @@ func (s *EnvironmentControllerTestSuite) SetupTest() {
 	s.router = NewRouter(nil, s.manager)
 }
 
+func (s *EnvironmentControllerTestSuite) TestList() {
```
MrSerth, line 36: If I understand it correctly, you have three cases here:

1. Returning an empty JSON without any execution environment
2. The `fetch` parameter with an expected value of true and an assertion to be called
3. The `fetch` parameter with an unexpected parameter that should not call the method.

What about the "happy path" of returning a list of cached execution environments? Or is that in the `TestNomadEnvironmentManager_List`?

internal/api/environments_test.go
```diff
@@ -32,6 +33,140 @@ func (s *EnvironmentControllerTestSuite) SetupTest() {
...
+		request, err := http.NewRequest(http.MethodGet, path.String(), nil)
+		s.Require().NoError(err)
+
+		s.router.ServeHTTP(recorder, request)
+		s.Equal(http.StatusBadRequest, recorder.Code)
+		s.manager.AssertNotCalled(s.T(), "List")
+	})
+}
+
+func (s *EnvironmentControllerTestSuite) TestGet() {
```
MrSerth, line 88: Similar to above: Is the happy path (known environment, just without the `fetch` parameter) not interesting or already covered (e.g., in the `TestNomadEnvironmentManager_Get`)? A simple "Yes" is okay 🙂

internal/environment/manager_test.go
```diff
@@ -98,6 +101,144 @@ func TestNewNomadEnvironmentManager(t *testing.T) {
...
+		assert.NoError(t, err)
+		assert.Equal(t, 1, len(environments))
+		assert.Equal(t, fetchedEnvironment, environments[0])
+	})
+}
+
+func mockWatchAllocations(apiMock *nomad.ExecutorAPIMock) {
+	call := apiMock.On("WatchAllocations", mock.Anything, mock.Anything, mock.Anything)
+	call.Run(func(args mock.Arguments) {
+		<-time.After(10 * time.Minute) // 10 minutes is the default test timeout
```
MrSerth, line 237: Nice comment, thanks!

internal/api/environments_test.go
MrSerth: Sure, happy to do that at any time 😜. I am glad you took the time to add these and I wish I could do the same.

Let's ship it! :shipit:

MrSerth (approved):

PR #15, opened 2021-10-21
Exclude IntelliJ and GoLand files

MrSerth:


mpass99 (approved):
We had talked about this earlier in the group and came to the conclusion that everyone should exclude the IDE files of their preferred IDE in the global gitignore.
(`git config --global core.excludesFile '~/.gitignore'`:
`**/.vscode/*`
`**/.idea/* `)
But since we all use the same IDE here, we can do it differently :man_shrugging:

PR #19, opened 2021-10-21
Create a new runner async after one gets claimed

mpass99:
When we claim a new runner, we make ~~two~~ **three** synchronous requests to Nomad.
- In "Sample" we create a new runner at the same time. We should do this asynchronously, as the error case is only logged anyway.
- In "Mark Runner As Used" we tell Nomad that we are using the runner (only for the recovery, in case Poseidon has to restart). We could also consider making this request asynchronous, but then we would also have to think about a concept for the error case of the asynchronous request.

Both requests need about 200ms each in the staging environment. (Unfortunately, there are no delays detectable in the local environment).

### DIFF
--- a/tests/e2e/environments_test.go
+++ b/tests/e2e/environments_test.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"strings"
 	"testing"
+	"time"
 )
 
 func TestCreateOrUpdateEnvironment(t *testing.T) {
@@ -114,6 +115,7 @@ func TestListEnvironments(t *testing.T) {
 		job.Name = &jobID
 		_, _, err := nomadClient.Jobs().Register(job, nil)
 		require.NoError(t, err)
+		<-time.After(tests.ShortTimeout) // Nomad needs a bit to create the job
 
 		// List without fetch should not include the added environment
 		response, err := http.Get(path) //nolint:gosec // because we build this path right above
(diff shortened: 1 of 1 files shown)


mpass99:
Yes, you have reflected the problem well. I like the idea of trying to delete the runner if the marking does not work.
The actions of the second paragraph of the "mark as used" case are correct, but I don't think the probability of occurrence is worth the complexity introduced.

One thing should not go unmentioned at this point: We only make the effort to mark all provided runners so that we don't have to fill up all the prewarming pools when we restart Poseidon.
We have already had [this discussion](https://gitlab.hpi.de/codeocean/codemoon/wiki/-/blob/main/docs/CodiMD/2021-05-07%20Interesting%20discussions%20and%20decisions.md#on-storing-the-used-state-of-runners-in-nomad-2021-06-09), so please stop me. But :wink:, we add an extra 200ms of latency (or extra complexity with asynchronous calls) and load generation at Nomad for an order of magnitude 3000 runner requests per day, in exchange for not having to start an order of magnitude smaller number of runners after a rare restart of Poseidon?

mpass99:
Or let's calculate this concretely... How big should the prewarming pool actually be?
Even on the day of the heaviest use, we had no more than 500 simultaneous users. Depending on usage and how long we set the container reuse time window, each user will request 1-4 (hereafter 2) containers in that hour. (500 * 2) containers per hour = 0.3 (say 1) containers per second. At this rate, Nomad takes less than one second on average to start the container. According to [Little's Law](https://en.wikipedia.org/wiki/Little%27s_law), we would need a prewarming pool size of one container.
Of course, in a worst case scenario, all 500 learners could ask for a container at the same time, so we would need such a large prewarming pool.
In several tests launching 500 containers in parallel, Nomad took 1 minute +- 4s in the staging environment.
So in summary, to save up to a minute in restarting Poseidon, we put up with about 6000 more requests per day and more Code complexity.

PR #28, opened 2021-11-03
Close evaluation stream for Nomad Job creation

### LINKED ISSUE #22: Synchronizing execution environments might not work
I noticed some strange errors when synchronizing (all) execution environments a few times repeatedly. In CodeOcean, I see an error: 

Bildschirmfoto 2021-10-29 um 09 40 48

In my tests, this error is permanent as soon as it occurred once and thus should be fixed.

In Poseidon, I see the following log output:

`
2021/10/29 09:36:11 http: superfluous response.WriteHeader call from github.com/openHPI/poseidon/pkg/logging.(*loggingResponseWriter).WriteHeader (logging.go:50)
`

In Ruby, the following line is logged:

`
Could not create execution environment in Poseidon, got response: {"status"=>500, "body"=>"{\"message\":\"error registering template job in API: couldn't register job: error registering Nomad job: Put \\\"http://localhost:4646/v1/jobs?namespace=default\\\": EOF\",\"errorCode\":\"UNKNOWN\"}", "response_headers"=>{"content-type"=>"application/json", "date"=>"Fri, 29 Oct 2021 07:36:11 GMT", "content-length"=>"187"}}
`

mpass99:
when set event handler have been finished.
This partially fixes #22.

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -136,7 +136,9 @@ func (a *APIClient) LoadRunnerJobs(environmentID dto.EnvironmentID) ([]*nomadApi
 	return jobs, occurredError
 }
 
-func (a *APIClient) MonitorEvaluation(evaluationID string, ctx context.Context) error {
+func (a *APIClient) MonitorEvaluation(evaluationID string, outerContext context.Context) error {
+	ctx, cancel := context.WithCancel(outerContext)
+	defer cancel()
 	stream, err := a.apiQuerier.EvaluationStream(evaluationID, ctx)
 	if err != nil {
 		return fmt.Errorf("failed retrieving evaluation stream: %w", err)
--- a/internal/nomad/nomad_test.go
+++ b/internal/nomad/nomad_test.go
@@ -168,7 +168,8 @@ func asynchronouslyMonitorEvaluation(stream chan *nomadApi.Events) chan error {
 	readOnlyStream := func() <-chan *nomadApi.Events { return stream }()
 
 	apiMock := &apiQuerierMock{}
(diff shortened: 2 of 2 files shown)


mpass99:
We should think about removing the unlimited "http_max_conns_per_client" configuration as this is an indicator for a Bug in Poseidon and "[Nomad uses documented reasonable defaults](https://learn.hashicorp.com/tutorials/nomad/production-deployment-guide-vm-with-consul)"

MrSerth (commented):
Thanks for tackling this issue. I looks good (codewise) but I was still experiencing issues. With all three systems (Nomad, Poseidon, CodeOcean) just restarted, I have the following log output:

[CodeOcean.log](https://github.com/openHPI/poseidon/files/7468709/CodeOcean.log)

[Poseidon.log](https://github.com/openHPI/poseidon/files/7468711/Poseidon.log)

[Nomad.log](https://github.com/openHPI/poseidon/files/7468710/Nomad.log)

As a result, Poseidon is still busy waiting using about 3.5 cores

Bildschirmfoto 2021-11-03 um 16 24 13


Sorry to ask again for your support here. And of course, I used a build generated from 55d231c9.

MrSerth:
> We should think about removing the unlimited "http_max_conns_per_client" configuration as this is an indicator for a Bug in Poseidon and "[Nomad uses documented reasonable defaults](https://learn.hashicorp.com/tutorials/nomad/production-deployment-guide-vm-with-consul)"

Yes, I would do so for testing purposes as well. We just might want to increase the value if we need to handle more than 100 requests per second (because each execution would also add a new [WebSocket] connection).

mpass99:
Obsolete after #31

PR #31, opened 2021-11-09
Infinite busy waiting for lost event

mpass99:
Thanks for finding [this bug](https://github.com/openHPI/poseidon/pull/28#pullrequestreview-796723048).

## *Who* is the bug affecting?
- The administrator, as the bug occurs particularly frequently in high load situations such as the "Synchronize all environments" functionality.
- Any user, as the error can always occur during heavy use.

## *What* is affected by this bug?
The complete usability of Poseidon, because with each occurrence of the bug a thread waits permanently busy and so the resources of the server are quickly overloaded.

## *When* does this occur?
The error occurs when Nomad throws the following error:
`2021-11-09T14:21:27.986Z [ERROR] http: request failed: method=GET path=/v1/event/stream?index=0&namespace=default&topic=Evaluation%3Ab5409340-3e32-c9de-b832-bff225d38983 error="event dropped from buffer" code=500`.

Poseidon does not realise this and so the existing error handling cannot take effect.

## *Where* on the platform does it happen?
In the local environment, the bug appears quickly - approximately every 25 runners to be created in one second.
In the staging environment, the error occurs much less frequently when an environment is updated. This is mainly due to the fact that the requests for updating each runner are synchronous and the RTT equalises the load for Nomad (700ms vs. 15ms).

## *How* do we replicate the issue?
- Use Postman to call the PUT Environment route
- Update an environment with about 50 prewarmed runner
- (Staging: Repeat the previous step some times)

## Expected behavior (i.e. solution)
It would be desirable to avoid or deal with Nomad's error. Exemplary solutions for both variants are indicated in this pull request.
- Deal: Introduce a timeout for all synchronous Nomad operations.
- Avoid: As we can see from Nomad's error message, the error has to do with the event stream that Poseidon requests. Every time Poseidon tasks Nomad with something, Poseidon requests in extra requests to get all updates on the task. This is an additional load that eventually leads to buffer overflow at Nomad.
Yet the extra event stream is unnecessary. Poseidon has a permanent event stream anyway, which watches out for Nomads allocations and transfers this information to the Poseidon data when creating / removing one. We can use this already existing stream to also pay attention to the other Nomad feedbacks such as "Error No more memory" or "Job created successfully".
The advantage of this is that we save many requests to Nomad and thus reduce the load. Thus, with a prototype-like solution, Nomad's buffer no longer ran full and thus all events could be received.
The disadvantage of this is that we would have to implement an internal waiting system for events so that possible errors can also be returned in the HTTP requests.


MrSerth:
These are really good findings and you created a thoughtful analysis of the things that happen at Nomad. From an architectural point of view, the possible solution you outlined in the "avoid" section (e.g. using the same stream to wait for the result of various tasks) feels more natural. However, I am unsure about the expense this might introduce. It sounds like a typical use case for a messaging queue that many listeners can attach to. Another approach would be to register callbacks that get executed once a success or failure message is received through the stream. Still, the problem you mentioned remains: The former synchronous request is now blocked by messages in the event stream and some form of notifications are required to allow Poseidon to return a HTTP response to the caller.

mpass99:
Yes, I would go with your second solution. Or, I was thinking of generating a simple dictionary that maps from the evaluation id to a Go channel. At the point where currently some events are processed, only the content of the events would have to be written into the correct channel.
In comparison, functionality would not be passed around, but would remain in one place. In addition, much of the existing code could be adopted and we would use a construct with the channels that is centrally integrated and optimised in Go. Therefore, I would classify the effort as relatively low.
Even if the probability of an event disappearing is lower with this variant, I would also leave the timeout in. This way we would have no leaking routines and guaranteed HTTP responses.

MrSerth:
> Yes, I would go with your second solution. Or, I was thinking of generating a simple dictionary that maps from the evaluation id to a Go channel. At the point where currently some events are processed, only the content of the events would have to be written into the correct channel.

This is totally fine for me; a simple yet effective method should be sufficient to deal with this issue.

> In comparison, functionality would not be passed around, but would remain in one place. In addition, much of the existing code could be adopted and we would use a construct with the channels that is centrally integrated and optimised in Go. Therefore, I would classify the effort as relatively low.

Sure, async processing is one of the advantages of Go and another argument to go this way.

> Even if the probability of an event disappearing is lower with this variant, I would also leave the timeout in. This way we would have no leaking routines and guaranteed HTTP responses.

I also agree here: We should always try to render a response and clean up any failures. This will allow a long running Poseidon without frequent needs to restart the service (allowing higher overall availability) 🙂

mpass99:
Closes #28

MrSerth (approved):
LGTM :shipit:

PR #34, opened 2021-11-10
Update versions used during GitHub actions

MrSerth:
This MR will use the latest versions for Go and Nomad in our GitHub actions pipeline.

PR #36, opened 2021-11-10
Upgrade dependencies

MrSerth:
This MR is based on the following command executions:

```
make upgrade-deps
make tidy-deps
```

It updates the dependencies to the latest version supported in the project (and also generates a new cache for GitHub actions).


MrSerth:
We still cannot update hcl2 to the latest version. Nomad uses a specialized hcl2 version that we must use as well:

https://github.com/openHPI/poseidon/blob/ac6ce56c38922d561c55814f6d2c23facb8491e2/go.mod#L42

MrSerth:
This PR is based on #34 to resolve failures in the e2e tests.

mpass99 (approved):

PR #37, opened 2021-11-11
Checkout code for trivy analysis

MrSerth:
This PR checks out the code in order to perform the privy analysis. Without our code, the analysis doesn't make much sense, right?


MrSerth:
The analysis now works as expected. Privy reports a bad status if new alerts are found. 

As of now, there are some alerts of our dependencies. I would still argue that we should merge this PR and observe the effects on other pull requests. I don't expect an error state to be reported if no new (but only existing) alerts are found.

mpass99 (approved):
LGTM

PR #39, opened 2021-11-17
Make environment variables of the system accessible.

### LINKED ISSUE #27: Environment variables from Dockerfile are not available
As an administrator of CodeOcean
I want to specify environment variables using the `Dockerfile`
in order to have a central interface for defining everything related to my executions

<hr>

In my tests, I noticed some differences in the handling of environment variables between the DockerContainerPool and Poseidon. With Poseidon, the environment variables defined in my `Dockerfile` are not set during command execution. For example, executing `env` in a container and through CodeOcean / Poseidon reveals a major difference (see screenshot).

We need to find a way that allows me to use the variables as defined in my `Dockerfile`. As they might be different per execution environment, they should be used exactly as defined in the image.

Bildschirmfoto 2021-11-03 um 00 14 27

mpass99:
Closes #27

The simplest solution for Issue #27 would be to make all environment variables of the system available (as implemented in this merge request).
The question remains, however, whether this is always the intention. So far, Poseidon's approach has been to make only those environment variables accessible that are declared in the execute request.
A way in between would be that only if the environment variable, e.g. "showSystemEnvironmentVariables" (or an HTTP query parameter) is set to "true", all system variables are available.

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -278,7 +278,7 @@ paths:
                   type: string
                   example: python exercise.py
                 environment:
-                  description: Environment variables for this execution. The keys of this object are the variable names and the value of each key is the value of the variable with the same name
+                  description: Environment variables for this execution. The keys of this object are the variable names and the value of each key is the value of the variable with the same name. The environment variables of the system remain accessible.
                   type: object
                   additionalProperties:
                     type: string
--- a/pkg/dto/dto.go
+++ b/pkg/dto/dto.go
@@ -24,7 +24,7 @@ type ExecutionRequest struct {
 
 func (er *ExecutionRequest) FullCommand() []string {
 	command := make([]string, 0)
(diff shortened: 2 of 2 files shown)


MrSerth:
Thanks for this fix, I was able to use Poseidon and Nomad successfully for my Java exercise.

When further investigating, I was able to list all environment variables. Most of them are intended and necessary (even when inherited from the system). I don't expect them to cause further (security) issues. The only variables I am not sure about (but was already well informed about their existence) are the ones prefixed with `NOMAD_`. They allow collecting some system information (but nothing too secret?) and we would not need them. Maybe we can hide only those? In any case, I would propose to create a new ticket and use the approach you suggested with this PR for now.

Nomad env

MrSerth (approved):

MrSerth:
I added issue #41 as a reminder unblocking a merge here.

PR #47, opened 2021-11-24
Add route for version information

### LINKED ISSUE #25: Expose release information through REST API
As an administrator on CodeOcean
I would like to get an overview of the system components
in order to check which version is currently running as part of my application.

<hr>

We should include a release identifier to expose it through a REST API. We might want to use a Git hash (as this makes tracking very easy) but if this is not feasible, we can also use something else (e.g. build time of the server). We might also investigate #26 as Sentry has some support for release information and is also required for proper error tracking.

mpass99:
Closes #25

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -106,6 +106,20 @@ paths:
         "204":
           description: Everything okay
 
+  /version:
+    get:
+      summary: Retrieve the version of Poseidon
+      description: Return hash-like release information.
+      tags:
+        - miscellaneous
+      responses:
+        "200":
+          description: The release information could be returned.
+        "404":
+          $ref: "#/components/responses/NotFound"
+        "500":
+          $ref: "#/components/responses/InternalServerError"
+
   /runners:
     post:
       summary: Provide a runner
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -3,6 +3,7 @@ package api
 import (
 	"github.com/gorilla/mux"
 	"github.com/openHPI/poseidon/internal/api/auth"
+	"github.com/openHPI/poseidon/internal/config"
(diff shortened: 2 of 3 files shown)


internal/api/api.go
```diff
@@ -59,3 +62,14 @@ func configureV1Router(router *mux.Router, runnerManager runner.Manager, environ
 		configureRoutes(v1)
 	}
 }
+
+// Version handles the version route.
+// It responds the release information stored in the configuration.
```
MrSerth, line 67: Do we need to add the version as an additional key-value-pair to the example configuration?

api/swagger.yaml
```diff
@@ -106,6 +106,20 @@ paths:
...
 
+  /version:
+    get:
+      summary: Retrieve the version of Poseidon
+      description: Return hash-like release information.
+      tags:
+        - miscellaneous
+      responses:
+        "200":
+          description: Everything okay
```
MrSerth, line 117: We might want to consider another description here :)

MrSerth (commented):
Looks good to me and is a very clean solution 👍

internal/api/api.go
mpass99: It is done automatically by our deployment playbook [cf704180b2c96118ad835a2805eb02f0029c91d0](https://gitlab.hpi.de/maximilian.pass/poseidon-deployment/-/commit/cf704180b2c96118ad835a2805eb02f0029c91d0)
But I now added a comment for that in the description. Thank you for pointing out this ambiguity.

https://github.com/openHPI/poseidon/blob/10754d778d83cd11403053e58797678c89c3ff0a/configuration.example.yaml#L56-L58

MrSerth: That's awesome, thanks! 🙏

Issue #52, opened 2021-11-26
Improve reliability of test pipeline


MrSerth:
We should improve our CI to reduce the number of false-positive failures. Mostly, the e2e tests are flaky and fail randomly, but sometimes the (unit) tests are also failing. See [previous attempts of this test run](https://github.com/openHPI/poseidon/actions/runs/1508904639) for more details

mpass99:
- 2 failures are due to the flaky test TestExecuteInteractivelyTestSuite/TestSendsSignalAfterTimeout
- 6 failures are due to the e2e tests - but not due to flaky test cases but due to a "socket hang up" error

PR #61, opened 2021-12-04
Add forcePull option

### LINKED ISSUE #33: Update Docker Image / Optimize Runner Update
As part of #22, we discussed optimization strategies for Runner updates. One idea was to skip updating runners if nothing changed (which in general is a good idea). For most values (e.g. the RAM oder CPU shares), this would work and reduce the load on Nomad. One part we weren't sure about was updating the Docker image.

In my tests, I found out that the Docker image will not be updated automatically (in Nomad) when updating the runner though Poseidon. According to the Nomad API, images with a specific tag (i.e. not `latest`) are not updated automatically. If all images should be updated, the [`force_pull`](https://www.nomadproject.io/docs/drivers/docker#force_pull) option in the task configuration could be used.

This option `force_pull = true` would be added in the config of the job:
https://github.com/openHPI/poseidon/blob/fff67246d6cc0bad92f403f55dd77dc5bcf62b6c/internal/environment/template-environment-job.hcl#L31-L36

When doing so, Nomad will pull the latest image for the tag provided and recreate environments when the image has changed. This is probably something we want to implement; I tested it successfully on my local machine.

However, when we rely on this Nomad feature, we probably cannot optimize the update process by skipping runner updates if nothing changed (the image could have been updated under the same name and thus no change is visible). A possible optimization could be to enable [rolling updates](https://learn.hashicorp.com/tutorials/nomad/job-rolling-update) which seem quite easy and would reduce the workload we put on Nomad when updating an environment with a large pool size.

mpass99:
Closes #33 
[...] for pulling the image when the execution environment gets updated.

### DIFF
--- a/docs/resources/client.example.hcl
+++ b/docs/resources/client.example.hcl
@@ -6,10 +6,14 @@ client {
     ]
 }
 
-# plugin "docker" {
-#   config {
-#     auth {
-#       config = "/root/.docker/config.json"
-#     }
-#   }
-# }
+plugin "docker" {
+    config {
+        gc {
+            image_delay = "1m"
+        }
+
+        # auth {
+            # config = "/root/.docker/config.json"
+        # }
+    }
+}
--- a/docs/resources/client.example.hcl
+++ b/docs/resources/client.example.hcl
@@ -9,7 +9,7 @@ client {
 plugin "docker" {
     config {
         gc {
-            image_delay = "1m"
+            image_delay = "0s"
         }
 
         # auth {
--- a/internal/runner/manager.go
+++ b/internal/runner/manager.go
@@ -51,9 +51,11 @@ type ExecutionEnvironment interface {
 	// Delete removes this environment and all it's runner from the executor and Poseidon itself.
 	Delete(apiClient nomad.ExecutorAPI) error
 	// Scale manages if the executor has enough idle runner according to the PrewarmingPoolSize.
-	Scale(apiClient nomad.ExecutorAPI) error
+	// With forcePull the current image is pulled again.
+	Scale(apiClient nomad.ExecutorAPI, forcePull bool) error
 	// UpdateRunnerSpecs updates all Runner of the passed environment to have the same definition as the environment.
-	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI) error
+	// With forcePull the current image is pulled again.
+	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI, forcePull bool) error
 
 	// Sample returns and removes an arbitrary available runner.
 	// ok is true iff a runner was returned.
@@ -217,7 +219,7 @@ func (m *NomadRunnerManager) Load() {
 		for _, job := range runnerJobs {
 			m.loadSingleJob(job, environmentLogger, environment)
 		}
-		err = environment.Scale(m.apiClient)
+		err = environment.Scale(m.apiClient, false)
 		if err != nil {
 			environmentLogger.WithError(err).Error("Couldn't scale environment")
 		}
--- a/internal/runner/manager.go
+++ b/internal/runner/manager.go
@@ -51,11 +51,9 @@ type ExecutionEnvironment interface {
 	// Delete removes this environment and all it's runner from the executor and Poseidon itself.
 	Delete(apiClient nomad.ExecutorAPI) error
 	// Scale manages if the executor has enough idle runner according to the PrewarmingPoolSize.
-	// With forcePull the current image is downloaded again from the registry.
-	Scale(apiClient nomad.ExecutorAPI, forcePull bool) error
+	Scale(apiClient nomad.ExecutorAPI) error
 	// UpdateRunnerSpecs updates all Runner of the passed environment to have the same definition as the environment.
-	// With forcePull the current image is downloaded again from the registry.
-	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI, forcePull bool) error
+	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI) error
 
 	// Sample returns and removes an arbitrary available runner.
 	// ok is true iff a runner was returned.
@@ -219,7 +217,7 @@ func (m *NomadRunnerManager) Load() {
 		for _, job := range runnerJobs {
 			m.loadSingleJob(job, environmentLogger, environment)
 		}
-		err = environment.Scale(m.apiClient, false)
+		err = environment.Scale(m.apiClient)
 		if err != nil {
 			environmentLogger.WithError(err).Error("Couldn't scale environment")
 		}
--- a/internal/runner/manager.go
+++ b/internal/runner/manager.go
@@ -51,10 +51,10 @@ type ExecutionEnvironment interface {
 	// Delete removes this environment and all it's runner from the executor and Poseidon itself.
 	Delete(apiClient nomad.ExecutorAPI) error
 	// Scale manages if the executor has enough idle runner according to the PrewarmingPoolSize.
-	// With forcePull the current image is pulled again.
+	// With forcePull the current image is downloaded again from the registry.
 	Scale(apiClient nomad.ExecutorAPI, forcePull bool) error
 	// UpdateRunnerSpecs updates all Runner of the passed environment to have the same definition as the environment.
-	// With forcePull the current image is pulled again.
+	// With forcePull the current image is downloaded again from the registry.
 	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI, forcePull bool) error
 
 	// Sample returns and removes an arbitrary available runner.
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -173,6 +173,7 @@ func (n *NomadEnvironment) SetNetworkAccess(allow bool, exposedPorts []uint16) {
 // Register creates a Nomad job based on the default job configuration and the given parameters.
 // It registers the job with Nomad and waits until the registration completes.
 func (n *NomadEnvironment) Register(apiClient nomad.ExecutorAPI) error {
+	nomad.SetForcePullFlag(n.job, true) // This must be the default as otherwise new runners could have different images.
 	evalID, err := apiClient.RegisterNomadJob(n.job)
 	if err != nil {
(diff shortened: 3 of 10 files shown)


docs/resources/client.example.hcl
```diff
@@ -6,10 +6,14 @@ client {
...
-#   config {
-#     auth {
-#       config = "/root/.docker/config.json"
-#     }
-#   }
-# }
+plugin "docker" {
+    config {
+        gc {
+            image_delay = "1m"
```
mpass99, line 12: > Then, the allocations will first redownload the image and finally start the runners. A potential side effect of this mechanism would be that images get updated / downloaded more often, depending on Nomad's scheduler.

What would be an appropriate timeout?

internal/runner/manager.go
```diff
@@ -51,9 +51,11 @@ type ExecutionEnvironment interface {
 	// Delete removes this environment and all it's runner from the executor and Poseidon itself.
 	Delete(apiClient nomad.ExecutorAPI) error
 	// Scale manages if the executor has enough idle runner according to the PrewarmingPoolSize.
-	Scale(apiClient nomad.ExecutorAPI) error
+	// With forcePull the current image is pulled again.
```
MrSerth, line 54:
```suggestion
	// With forcePull the current image is downloaded again from the registry.
```

internal/runner/manager.go
```diff
@@ -51,9 +51,11 @@ type ExecutionEnvironment interface {
...
+	Scale(apiClient nomad.ExecutorAPI, forcePull bool) error
 	// UpdateRunnerSpecs updates all Runner of the passed environment to have the same definition as the environment.
-	UpdateRunnerSpecs(apiClient nomad.ExecutorAPI) error
+	// With forcePull the current image is pulled again.
```
MrSerth, line 57:
```suggestion
	// With forcePull the current image is downloaded again from the registry.
```

docs/resources/client.example.hcl
MrSerth: Good question; I don't know. However, I would consider the following scenario to determine the delay time:

Let's assume we have two agents (A and B) running allocations of our execution environment and the system is being used actively. First, both agents have some allocations with image version 1. Due to the inactivity timeout, the last allocation is stopped on agent A (all other allocations are on agent B). Within the `image_delay`, I instruct Nomad to download a newer image version 2 and recreate the environments. I see two possibilities here:

1. The allocations on agent B get stopped and are recreated on agent B exclusively. I assume that all allocations will have the new image version 2.
2. The allocations on agent B get stopped and are recreated on agents A and B. As the `force_pull` option is included for all allocations, all allocations should use the new image 2.

This is no issue so far. However, what happens if we continue the experiment and assume the first outcome (all allocations updated on agent B). Now, a learner requests a new runner, and one from agent B is taken (as agent A doesn't have an allocation yet). Still, while being in the `image_delay`, Poseidon creates a new allocation to refill the prewarming pool size. This time, the allocation is created on agent A (due to Nomad's internal scheduler). 

My question: Which image version is used by the new allocation?

MrSerth (commented):
Thanks, looks good so far. I mainly have one question regarding the delay time and searched for support of named arguments in Go (which unfortunately is not supported). Generally speaking, I am mostly not a big fan of unnamed boolean arguments to methods because their intended use case is not immediately visible:

```go
n.createRunner(apiClient, false)
```

Instead, I would have preferred a self-documenting mechanism, such as 

```go
n.createRunner(apiClient, forcePull: false)
```

The only mechanism in Go might be a map which I feel is too much for this use case here. So I would suggest keeping it as you did.

mpass99:
It would be image version 1.

And that would be a faulty behaviour. Whereby this behaviour is limited.
- In the case of a large usage, Nomad will distribute the runners equally among the workers. This would cause all Docker images to be updated by the force_pull option.
- Only the case of small usage (/number of idle and used runners) is critical. In this case, some workers have no runner of the updated environment. It depends strongly on the usage how likely it is that a new runner will be requested in the delay time.

One could eliminate this error case by tracking on the Poseidon side whether one is still in the delay time after updating an image, and then setting the force_pull option. Do you think we should do that?

mpass99:
> Generally speaking, I am mostly not a big fan of unnamed boolean arguments to methods because their intended use case is not immediately visible

I agree. I am a fan of Goland's way providing this information:
Parameter description

Makefile
```diff
@@ -95,6 +96,7 @@ e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
 
 .PHONY: e2e-test
 e2e-test: deps e2e-test-docker-image ## Run e2e tests
+	@docker image ls
 	@go test -count=1 ./tests/e2e -v -args -dockerImage="$(E2E_TEST_DOCKER_IMAGE)"
```
mpass99, line 100: This pull requests breaks our e2e ci pipeline. Currently we are building a local image from a Dockerfile. Then we create an environment with this image:
https://github.com/openHPI/poseidon/blob/85895e17075d23f7581a2647cf1da0edae48d37f/tests/e2e/e2e_test.go#L77-L88
But we just introduced the force_pull flag there so local images are ignored and so the image can not be found.

We could solve this issue for example by using an image from a registry instead of building our own.

MrSerth:
> It would be image version 1.
> 
> And that would be a faulty behaviour. Whereby this behaviour is limited.
> 
> * In the case of a large usage, Nomad will distribute the runners equally among the workers. This would cause all Docker images to be updated by the force_pull option.
> * Only the case of small usage (/number of idle and used runners) is critical. In this case, some workers have no runner of the updated environment. It depends strongly on the usage how likely it is that a new runner will be requested in the delay time.
> 
> One could eliminate this error case by tracking on the Poseidon side whether one is still in the delay time after updating an image, and then setting the force_pull option. Do you think we should do that?

Thanks for investigating and providing these insights! It doesn't surprise me but (of course) would have wished that the update is spread across all nodes.

I think we should prevent this error if possible. In my opinion, I see the following three options:

1. Always use `force_pull`. This might increase the start time of new containers in any load situation
2. Remove the [`image_delay`](https://www.nomadproject.io/docs/drivers/docker#image_delay) and remove containers immediately. This might increase the start time of new containers in low load situations
3. Track the delay in Poseidon and use the `force_pull` option during that time span. This might increase the complexity in Poseidon and (for the duration of the delay time) increase the start time of new containers in any load situation.
4. Don't reuse image tags. This would allow us to even increase the delay time and would require a version identifier for those images that get updated (frequently).

I tend towards option 2 or 4. Option 2 sounds like an acceptable (but sub-optimal) quick fix while option 4 seems to be more semantically correct.

Makefile
MrSerth: Ah, that's a good idea and would also increase the e2e execution time. What if we just create a Docker image, push it to this repository's registry, and use the (static) image in our tests?

I would recommend tracking that in another PR first (decoupled from this one), merging the other, and then just rebasing here.

mpass99:
Yes, I would agree going with option 2 or 4.
For option 2 I [verified](https://github.com/hashicorp/nomad/blob/2a8f0321720f8ec49b4b3bd0881114e396a599ad/drivers/docker/coordinator.go#L273-L278) that the docker image is only removed when no runner uses this image anymore.

I quickly implemented option 2 because the implementation here is trivial and we thus have a fixed reference point. For that option we still need to update the `client.hcl` in the staging and production environment.

MrSerth (approved):
Thanks for updating this PR and implementing option two. I applied the updated config option to our Nomad agents so that we're ready to deploy these changes once they (and the e2e test changes) were merged). :shipit:

MrSerth (approved):
Thanks for rebasing and finishing the e2e requirement. I am happy that (after eight failing test runs) the pipeline succeeds with the changes so that this should be ready 🧙

mpass99:
8 Failures. According to my current knowledge of Mathematics 3, this might be a high probability. Why? With the other pull requests, I rarely had to restart the CI...

2 failures are due to the flaky test TestExecuteInteractivelyTestSuite/TestSendsSignalAfterTimeout
6 failures are due to the e2e tests - but not due to flaky test cases but due to the fact that Nomad could not create an environment with 10 runners within 30 seconds. Maybe because it now re-pulls the image? I will observe this during another CI run

MrSerth:
> 8 Failures. According to my current knowledge of Mathematics 3, this might be a high probability. Why? With the other pull requests, I rarely had to restart the CI...
> 
> 2 failures are due to the flaky test TestExecuteInteractivelyTestSuite/TestSendsSignalAfterTimeout 6 failures are due to the e2e tests - but not due to flaky test cases but due to the fact that Nomad could not create an environment with 10 runners within 30 seconds. Maybe because it now re-pulls the image? I will observe this during another CI run

Thanks for checking and investigating. 30 seconds sounds like a lot even when the image is re-downloaded each time. Didn't you [mention](https://github.com/openHPI/poseidon/issues/33#issuecomment-979429763) that only the digest might get compared. Otherwise, I agree: It's really a high probability of failing tests.

mpass99:
Yes, as far as I know, only the digests are compared.
I was able to reproduce the problem in my environment. The e2e test gets an EOF response. Then I was able to recreate the error in Postman. The simple creation of the test execution environment fails.
When debugging, however, I was able to step through the "createOrUpdate" function without any problems. Just at the moment when the writer sends back the corresponding status message (line 129), Postman displays "Socket hang up". After a few repetitions, I changed the environment ID. From then on, I could no longer reproduce this error, whatever I did...

With this I can be relatively sure that the error is related to this pull request.

MrSerth:
That's really interesting. I tried following your explanation but was unsure about the line you referenced. Is it [this one](https://github.com/openhpi/poseidon/blob/34d4bb7ea0afd98ed56781c01d28ff447b151b4a/internal/api/environments.go#L129)? And I don't have any idea why this should be connected to the execution environment ID - that's not obvious, isn't it?

As you mention that this change here is _not_ related to the flaky tests: Shall we move #52 up and tackle the test reliability next? I would expect that more reliable tests make it easier for us to identify real errors and save us from debugging each PR individually.

mpass99:
> Is it this one?

Yes

> that's not obvious, isn't it?

I agree. Maybe it was just that I introduces any change in the request, or it was just random. After I changed it once I could not reproduce the error by returning to environment 0.

> tackle the test reliability next?

Ok, let's do that

PR #62, opened 2021-12-05
Add statistics route for execution environments

### LINKED ISSUE #24: Design and implement statistical route about pool size
As an administrator on CodeOcean  
I would like to get an overview of allocations/runner currently used or prewarmed
in order to get an impression of the overall system load.

<hr>

On CodeOcean, we have some statistics for each execution environment. We should design an API and implement it in Poseidon (and DCP afterward) to provide the information required. I assume the following information is useful per execution environment:

- Desired pre-warming pool size 
- Actual pre-warming pool size 
- Used runner pool size (as reserved for specific users and not available for new users)

This is how it the dashboard looks like right now (not all information included yet):

Bildschirmfoto 2021-11-01 um 14 07 43

mpass99:
Closes #24

Please push me if you see a use case for an additional test ;)

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -214,6 +214,7 @@ paths:
                           description: The port inside the container.
                           type: integer
                           minimum: 0
+                          maximum: 65535
                           example: 80
                         hostAddress:
                           description: The address which can be contacted to reach the mapped port.
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -120,6 +120,51 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /statistics/execution-environments:
+    get:
+      summary: Retrieve the statistics about the execution environments of Poseidon
+      description: Return Return the current availability and usage of runners.
+      tags:
+        - miscellaneous
+      responses:
+        "200":
+          description: Success. Returns all execution environments
+          content:
+            application/json:
+              schema:
+                type: object
+                additionalProperties:
+                  type: object
+                  properties:
+                    id:
+                      description: The id of the execution environment.
+                      type: integer
+                    prewarmingPoolSize:
+                      description: Number of runners with this configuration to prewarm.
+                      type: integer
+                      example: 50
+                    idleRunners:
+                      description: Number of runners currently prewarmed.
+                      type: number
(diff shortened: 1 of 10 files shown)


api/swagger.yaml
```diff
@@ -167,7 +212,7 @@ paths:
                       properties:
                         exposedPort:
                           description: The port inside the container.
-                          type: uint
+                          type: number
```
MrSerth, line 215: Is this an intended change here?

internal/api/api.go
```diff
@@ -73,3 +79,16 @@ func Version(writer http.ResponseWriter, _ *http.Request) {
 		writer.WriteHeader(http.StatusNotFound)
 	}
 }
+
+// StatisticsExecutionEnvironments handles the route for statistics about execution environments.
+// It responds the prewarming pool size and the amount of idle runners and used runners.
```
MrSerth, line 84: It's probably the "number" of runners as they are countable, isn't it?

MrSerth (commented):
Looks good so far :) I'll try to implement support for the new route in CodeOcean soon and provide another update here ;)

MrSerth (changes requested):
I implemented support for the new route in the following PR: https://github.com/openHPI/codeocean/pull/1105

While doing so, I noticed a small mistake. Otherwise, I like the new feature 🎉

api/swagger.yaml
mpass99: Intended yes, but not as correct as it should be

api/swagger.yaml
```diff
@@ -167,7 +212,8 @@ paths:
                       properties:
                         exposedPort:
                           description: The port inside the container.
-                          type: uint
+                          type: integer
+                          minimum: 0
```
MrSerth, line 216:
```suggestion
                          minimum: 0
                          maximum: 65535
```

We might also specify the maximum port number but don't need to.

MrSerth (approved):
Thanks for the (small) fix in the name of the JSON attributes. This will now work as expected for CodeOcean 🙂

api/swagger.yaml
MrSerth: Ah, I see. Thanks for pointing that out!

PR #63, opened 2021-12-07
Change the e2e test rule to use the remote image

mpass99:
Resolves: https://github.com/openHPI/poseidon/pull/61#discussion_r762589298

The e2e test can not work until #64 is merged to the main branch.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -98,6 +98,7 @@ e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
 
 .PHONY: e2e-test
 e2e-test: deps ## Run e2e tests
+	@docker pull $(E2E_TEST_DOCKER_IMAGE)
 	@go test -count=1 ./tests/e2e -v -args -dockerImage="$(E2E_TEST_DOCKER_IMAGE)"
 
 .PHONY: e2e-docker
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,15 @@
 PROJECT_NAME := "poseidon"
-PKG := "github.com/openHPI/$(PROJECT_NAME)/cmd/$(PROJECT_NAME)"
+REPOSITORY_OWNER = "openHPI"
+PKG := "github.com/$(REPOSITORY_OWNER)/$(PROJECT_NAME)/cmd/$(PROJECT_NAME)"
 UNIT_TESTS = $(shell go list ./... | grep -v /e2e)
 
 DOCKER_E2E_CONTAINER_NAME := "$(PROJECT_NAME)-e2e-tests"
 DOCKER_TAG := "poseidon:latest"
 DOCKER_OPTS := -v $(shell pwd)/configuration.yaml:/configuration.yaml
-# don't use :latest to prevent Nomad from trying to pull the image automatically
-E2E_TEST_DOCKER_IMAGE = "poseidon/e2e-docker-image:ci"
+LOWER_REPOSITORY_OWNER = $(shell echo $(REPOSITORY_OWNER) | tr A-Z a-z)
+REGISTRY = "ghcr.io"
+TAG = ":main"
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -89,12 +89,8 @@ coverage: deps ## Generate code coverage report
 coverhtml: coverage ## Generate HTML coverage report
 	@go tool cover -html=coverage_cleaned.cov -o coverage_unit.html
 
-.PHONY: e2e-test-docker-image ## Build Docker image used in e2e tests
-e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
-	@docker build -t $(E2E_TEST_DOCKER_IMAGE) deploy/e2e-test-image
-
```
MrSerth, line 95: Should we keep these lines to keep the "knowledge" about building the docker image? In theory, I should be able to build the image locally and push it to the registry, right? I am unsure, we might also drop it (the line is not hard to remember and basic Docker knowledge) and local e2e tests will also fetch the remote image.

MrSerth (approved):
Thanks, looks good. I am eager to see it live together with #64

Makefile
```diff
@@ -1,12 +1,15 @@
...
 UNIT_TESTS = $(shell go list ./... | grep -v /e2e)
 
 DOCKER_E2E_CONTAINER_NAME := "$(PROJECT_NAME)-e2e-tests"
 DOCKER_TAG := "poseidon:latest"
 DOCKER_OPTS := -v $(shell pwd)/configuration.yaml:/configuration.yaml
-# don't use :latest to prevent Nomad from trying to pull the image automatically
-E2E_TEST_DOCKER_IMAGE = "poseidon/e2e-docker-image:ci"
+LOWER_REPOSITORY_OWNER = $(shell echo $(REPOSITORY_OWNER) | tr A-Z a-z)
+REGISTRY = "ghcr.io"
+TAG = ":main"
```
mpass99, line 11: Is this tag ok or should we investigate how and why Github is automatically using this tag?

MrSerth: To be honest: I don't mind, that's fine for me (at least when it works as intended with the tests). I know that usually, Docker uses the `latest` tag but GitHub defaults to the branch name (which is `main`). 

From my perspective, we don't need to continue investigating that any further.

mpass99:
The docker pull takes longer with the remote registry than with an already pulled image. Therefore, the first tests fail because the image is not yet downloaded there and we are not waiting for it.
We might have to adapt our e2e test setup dynamically or statically to it.

MrSerth:
Ah, I see. Would it work to download the image beforehand with the `service` directive in the actions YAML? I do the same "trick" in the pipeline for CodeOcean. There, I instruct the CI to create a new service with the image. As a consequence, GitHub will download it, (spwan a container I don't require). See [these lines](https://github.com/openHPI/codeocean/blob/a4003ca7335c47272eca6cc15a2c042fb2882c0b/.github/workflows/ci.yml#L19-L20) for an example.

Makefile
```diff
@@ -98,6 +98,7 @@ e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
 
 .PHONY: e2e-test
 e2e-test: deps ## Run e2e tests
+	@docker pull $(E2E_TEST_DOCKER_IMAGE)
```
mpass99, line 101: Good trick! I have now implemented it the Makefile-way, as I find that the individual steps are then more together. But if you think that Github can optimise better if the itself hosted images are in the YAML, or something else, then I will adapt it that way.

MrSerth: That's true and I also like the close relationship between pulling the image and the usage in the tests 🙂

MrSerth (approved):
Looks good, thank you! And I absolutely like the smart solution of building the lowercase repo owner name automatically 😉

PR #66, opened 2021-12-08
Catch the client connection "Close normal" error

### LINKED ISSUE #49: Error when client disconnects
See [this Sentry issue](https://sentry.io/share/issue/662aa9f33d2548e1b4001e3793480fa3/)

mpass99:
Closes  #49

The loop that reads the stdin notes that the client has left the connection, just before our close handler does.
Now we catch this case.

### DIFF
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -76,7 +76,11 @@ func (cr *codeOceanToRawReader) readInputLoop(ctx context.Context) {
 		case <-readMessage:
 		}
 
-		if err != nil {
+		if err != nil && websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+			log.Debug("ReadInputLoop: The client closed the connection!")
+			// The close handler will do something soon.
+			return
+		} else if err != nil {
 			log.WithField("remote", cr.connection.(*websocket.Conn).UnderlyingConn().RemoteAddr()).
 				WithError(err).Warn("Error reading client message")
 			return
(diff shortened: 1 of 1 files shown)


internal/api/websocket.go
```diff
@@ -76,7 +76,11 @@ func (cr *codeOceanToRawReader) readInputLoop(ctx context.Context) {
 		case <-readMessage:
 		}
 
-		if err != nil {
+		if err != nil && websocket.IsCloseError(err, websocket.CloseNormalClosure) {
+			log.Debug("ReadInputLoop: The client closed the connection!")
+			// The close handler will do something soon.
```
MrSerth, line 81: We are experiencing a race condition here, right?

MrSerth (approved):
Pretty simple fix, thanks!

internal/api/websocket.go
mpass99: Yes

MrSerth: Alright. I think this handling is fine.

mpass99:
> Pretty simple fix, thanks!

Apart from the fact that this additional condition was exactly the mü for a too high cognitive complexity (for our linter), yeah...

MrSerth (approved):
> > Pretty simple fix, thanks!
> 
> Apart from the fact that this additional condition was exactly the mü for a too high cognitive complexity (for our linter), yeah...

Yes, I noticed that as well. Your solution is smart even when it took me a moment to get it (which is not related to the new version but rather the diff and the changes I compared).

PR #68, opened 2021-12-09
Improve test reliability

### LINKED ISSUE #52: Improve reliability of test pipeline

mpass99:
Closes #52

### DIFF
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -72,11 +72,10 @@ func initServer() *http.Server {
 	}
 
 	return &http.Server{
-		Addr:         config.Config.Server.URL().Host,
-		WriteTimeout: time.Second * 15,
-		ReadTimeout:  time.Second * 15,
-		IdleTimeout:  time.Second * 60,
-		Handler:      api.NewRouter(runnerManager, environmentManager),
+		Addr:        config.Config.Server.URL().Host,
+		ReadTimeout: time.Second * 15,
+		IdleTimeout: time.Second * 60,
+		Handler:     api.NewRouter(runnerManager, environmentManager),
 	}
 }
 
--- a/internal/runner/runner_test.go
+++ b/internal/runner/runner_test.go
@@ -175,10 +175,15 @@ func (s *ExecuteInteractivelyTestSuite) TestSendsSignalAfterTimeout() {
 	s.mockedExecuteCommandCall.Run(func(args mock.Arguments) {
 		stdin, ok := args.Get(4).(io.Reader)
 		s.Require().True(ok)
-		buffer := make([]byte, 1)                                                  //nolint:makezero,lll // If the length is zero, the Read call never reads anything. gofmt want this alignment.
-		for n := 0; !(n == 1 && buffer[0] == SIGQUIT); n, _ = stdin.Read(buffer) { //nolint:errcheck,lll // Read returns EOF errors but that is expected. This nolint makes the line too long.
+		buffer := make([]byte, 1) //nolint:makezero,lll // If the length is zero, the Read call never reads anything. gofmt want this alignment.
+		for n := 0; !(n == 1 && buffer[0] == SIGQUIT); {
 			time.After(tests.ShortTimeout)
+			n, _ = stdin.Read(buffer) //nolint:errcheck,lll // Read returns EOF errors but that is expected. This nolint makes the line too long.
+			if n > 0 {
+				log.WithField("buffer", fmt.Sprintf("%x", buffer[0])).Info("Received Stdin")
+			}
 		}
+		log.Info("After loop")
 		close(quit)
 	}).Return(0, nil)
 	timeLimit := 1
@@ -186,10 +191,12 @@ func (s *ExecuteInteractivelyTestSuite) TestSendsSignalAfterTimeout() {
 	s.runner.StoreExecution(defaultExecutionID, executionRequest)
 	_, _, err := s.runner.ExecuteInteractively(defaultExecutionID, bytes.NewBuffer(make([]byte, 1)), nil, nil)
 	s.Require().NoError(err)
+	log.Info("Before waiting")
 	select {
 	case <-time.After(2 * (time.Duration(timeLimit) * time.Second)):
 		s.FailNow("The execution should receive a SIGQUIT after the timeout")
 	case <-quit:
+		log.Info("Received quit")
 	}
 }
 
--- a/tests/e2e/websocket_test.go
+++ b/tests/e2e/websocket_test.go
@@ -63,9 +63,12 @@ func (s *E2ETestSuite) TestOutputToStdout() {
 	s.Require().Error(err)
 	s.Equal(&websocket.CloseError{Code: websocket.CloseNormalClosure}, err)
 
-	s.Require().Equal(&dto.WebSocketMessage{Type: dto.WebSocketMetaStart}, messages[0])
-	s.Require().Equal(&dto.WebSocketMessage{Type: dto.WebSocketOutputStdout, Data: "Hello World\r\n"}, messages[1])
-	s.Require().Equal(&dto.WebSocketMessage{Type: dto.WebSocketExit}, messages[2])
+	controlMessages := helpers.WebSocketControlMessages(messages)
+	s.Require().Equal(&dto.WebSocketMessage{Type: dto.WebSocketMetaStart}, controlMessages[0])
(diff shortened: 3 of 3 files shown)


mpass99:
**E2E Failure**
How it is sometimes:
``` log
time="2021-12-09T18:20:39.980373Z" level=info msg="Got Connection" ConnInfo="{0xc000114008 false false 0s}" package=test-helper
time="2021-12-09T18:21:10.118036Z" level=info msg="API: Just before sending response status" created=true package=api
time="2021-12-09T18:21:10.118289Z" level=fatal msg="Couldn't create default environment for e2e tests" error="error executing request: EOF" package=e2e
FAIL	github.com/openHPI/poseidon/tests/e2e	30.155s
```

How it should always be:
``` log
time="2021-12-09T18:13:24.264255Z" level=info msg="Got Connection" ConnInfo="{0xc0000107a8 false false 0s}" package=test-helper
time="2021-12-09T18:13:34.406509Z" level=info msg="API: Just before sending response status" created=true package=api
time="2021-12-09T18:13:34.406671Z" level=info msg="Got first response byte" package=test-helper
time="2021-12-09T18:13:34.406769Z" level=info msg="Put Idle Conn" error="<nil>" package=test-helper
ok  	github.com/openHPI/poseidon/tests/e2e	20.158s
```

Since we do not receive the first response byte in the error case I assume that the server http handler is crashing.
Site note: I wonder why in all 10 e2e failures I have observed, the execution time is exactly 20.1xx or 30.1xx seconds...

mpass99:
**E2E Failure**
This error is caused by our set `WriteTimeout` configuration.

https://github.com/openHPI/poseidon/blob/dce895faff56de11e8c59c03032e097aff3a7760/cmd/poseidon/main.go#L76

I cannot remember why we have introduced this. One reason I can think of is to interrupt a faulty state of Poseidon. In practice, however, this does not solve the problem, since the complete Poseidon functionality is calculated and only then is it decided whether the calculated result is to be returned or discarded..

I think we should remove this configuration.

mpass99:
**TestSendsSignalAfterTimeout**

The go race detector identifies a race condition in this test case. But this is intended and with one second it should be enough time to schedule the right goroutine..

Since I added debug statements I cannot reproduce the error for many tries. For me it would be fine to keep it this way and investigate further when it becomes a problem again.

MrSerth:
That's a really surprising result. As I was able to verify, this change dramatically increases the probability of a successful pipeline run. In my test, [only 8% of all E2E tests would still fail](https://github.com/openHPI/poseidon/actions/runs/1561017935), whereas with the write timeout [44% of all E2E tests would fail](https://github.com/openHPI/poseidon/actions/runs/1561034196). This is a huge difference.

While I am able to _observe_ the change, I am still unable to _understand_ it. According to [this](https://blog.cloudflare.com/the-complete-guide-to-golang-net-http-timeouts/) or [that](https://ieftimov.com/post/make-resilient-golang-net-http-servers-using-timeouts-deadlines-context-cancellation/) source, the write timeout is not seen as something negative but rather a protective feature for slow and untrusted clients. From what I've read, it _should_ work as a timeout and only cancel the request if it takes longer. Why is that not the case?

Additionally, I am wondering how this change might affect our production environment. For sure, we can assume CodeOcean is a trustworthy client (at least it should behave "normally"). I don't expect a major negative impact but couldn't it be that if a client is slow, we allocate too many resources in Poseidon? And why is this error not an issue for CodeOcean but only the E2E tests? If we found a real server error, I would have assumed that we could see similar errors in our production environment.

This change might be out of my horizon and I am fine if you acknowledge that this makes sense. I just want to prevent that we're fighting symptoms instead of the root cause...

mpass99:
Thank you for raising unclear points like this!

> From what I've read, it should work as a timeout and only cancel the request if it takes longer. Why is that not the case?

Approximately it is. In your first reference we find the practical definition of the `WriteTimeout`: `WriteTimeout normally covers the time from the end of the request header read to the end of the response write (a.k.a. the lifetime of the ServeHTTP)`

With this I am quiet sure our root cause is that the request to create a new environment with ten runners takes longer than 15s and so the response is timed out.

> a protective feature for slow and untrusted clients

I agree that both references warn about this vulnerability. Untrusted clients could read the response very slow and at the same time build up many connection leading to too many open connection (/reserved resources) and a crash of Poseidon. I see four solutions for this:
- Declare Poseidon clients as trusted and accept this vulnerability.
  - We are vulnerable for a DOS attack by all members of the same network.
- Approximate a maximum duration for all Poseidon requests and set it as `WrtiteTimeout`
  - ["While this is a timeout-like behavior, it would be more useful to stop our server from further execution when it reaches the timeout, ending the request. In our example above, the handler proceeds to process the request until it completes, although it takes 100% longer (2 seconds) than the response write timeout time (1 second)."](https://ieftimov.com/post/make-resilient-golang-net-http-servers-using-timeouts-deadlines-context-cancellation/)
- Introduce a [`LimitListener`](https://pkg.go.dev/golang.org/x/net/netutil#LimitListener) that provides a maximum number of simultaneous connections.
  - We would also be vulnerable to DOS as we block all new requests and do not end the malicious slow reading ones.
- Introduce a [`TimeoutHandler`](https://pkg.go.dev/net/http#TimeoutHandler) and set a timeout duration specific for each route.
  - Pro: We stop the execution in the moment the timeout is reached.
    Contra: We continue to limit our execution time. The creation of an execution environment with x runners is potentially aborted in the middle.

> And why is this error not an issue for CodeOcean but only the E2E tests?

Because usually CodeOcean runs only requests below 15s. And apart from that, it is. You have to try creating a new execution environment. When everything is correct except that it took longer than 15 seconds CodeOcean thows the error: `Docker Image error: Could not synchronize execution environment because of Faraday error: #<Faraday::SSLError wrapped=#<OpenSSL::SSL::SSLError: SSL_read: decryption failed or bad record mac>>`

> only 8% of all E2E tests would still fail

Thanks for pointing out that our familiar error of the lost line-break is still there.

mpass99:
**TestOutputToStdout**
What made the test flaky is that Nomad sometimes splits the stdout into multiple WebSocket messages. We often put these messages back together. In this test, however, we explicitly accessed a specific ws message.

MrSerth:
Now I get it, thanks for your explanation! I can relate to the SSL error (as I've seen this in CodeOcean before) and it reminds me of the synchronous response for new execution environments (which is good). I defentiley see the issue and also acknowledge that a response might take longer than 15 seconds in these cases.

I also think that the current approach of "terminating" the connection without sending a proper timeout response is not the best way to handle it: CodeOcean (or other clients) might fail with the TLS error and it causes the E2E tests to fail more often. Consequently, we should change it.

You presented several options to reintroduce the protective feature handled by the 15 seconds timeout - thanks for these! I came up with the similar ideas. All are valid options, based on the thread model that we face. For our use case, I would argue as follows: 

- Providing different timeouts for different routes would allow us to limit the attack to some routes but does not prevent it. Furthermore, any time limit could be reached by increasing the prewarming pool size. I think, it will shift the issue but not resolve it. Still, if we enable any `WriteTimeout`, the introduction of a `TimeoutHandler` would be suitable.
- `LimitListener` will not prevent slow connections from blocking the all listeners.
- Removing the timeout removes this protective feature at all (where 15 seconds was long anyway).

Hence, **I would follow your suggestion to remove the timeout**. Most other solutions will be too difficult to implement and maintain, I think. There is only one idea that came into my mind and mixes various concepts: Can we set a (relatively low) timeout for all unauthorized requests (e.g. those not containing the `Poseidon-Token`)? This will only affect the `/health` and any access to the API from unauthorized clients. As a consequence, an attacker is not able to exploit the service with slow connections if the token is not known (which would require any other attack to succeed beforehand). I consider even this idea as optional as we're operating an internal service anyway (even though it would increase our protection and thus might be a dedicated ticket?).  

Back to the E2E tests: I also like your second approach of combining the output and agree: If we do that on many other places, we should also do it here. In my [25 tests](https://github.com/openHPI/poseidon/actions/runs/1567121468), none of the E2E pipeline runs failed with these (well-known) errors. There was just a single failure of the `TestStderrFifoIsRemoved` which obviously does not occur that often.

MrSerth (approved):
Thanks for investing your time and effort into this, I am convinced it will pay off rather soon! 💰

Issue #44, opened 2021-11-18
Unexpected error received from Poseidon


MrSerth:
In Sentry, I noticed some reoccurring issues with errors received from Poseidon and tried to debug them. Below is an example that occurred today. It shows that the messages were received correctly and (from my expectation) are complete. However, instead of terminating the execution normally, Poseidon sends an error to CodeOcean.

Is there something we can improve here to avoid the error? Logs are attached for CodeOcean and Poseidon. As the connection to Nomad was closed normally, there might be some race condition?

CodeOcean:

```
D, [2021-11-18T14:57:09.389856 #3606] DEBUG -- : 2021-11-18 14:57:09.389779572 UTC: Receiving from wss://poseidon.codeocean.xopic.de:7200/api/v1/runners/29-98c10ec0-487f-11ec-b035-fa163e05d94e/websocket?executionID=26da8d47-b887-480a-aa09-31cfe79c23f7: "{\"type\":\"stdout\",\"data\":\"************* Module aufgabe123\\r\\n aufgabe123.py:1: convention (C0326, bad-whitespace, ) Exactly one space required after comma\\r\\n print(\\\"Viel\\\",\\\"Spaß\\\",\\\"im\\\",\\\"Sommercamp\\\")\\r\\n             ^\\r\\n aufgabe123.py:1: convention (C0326, bad-whitespace, ) Exactly one space required after comma\\r\\n print(\\\"Viel\\\",\\\"Spaß\\\",\\\"im\\\",\\\"Sommercamp\\\")\\r\\n                    ^\\r\\n aufgabe123.py:1: convention (C0326, bad-whitespace, ) Exactly one space required after comma\\r\\n print(\\\"Viel\\\",\\\"Spaß\\\",\\\"im\\\",\\\"Sommercamp\\\")\\r\\n                         ^\\r\\n \\r\\n\"}"
D, [2021-11-18T14:57:09.390751 #3606] DEBUG -- : 2021-11-18 14:57:09.390718572 UTC: Receiving from wss://poseidon.codeocean.xopic.de:7200/api/v1/runners/29-98c10ec0-487f-11ec-b035-fa163e05d94e/websocket?executionID=26da8d47-b887-480a-aa09-31cfe79c23f7: "{\"type\":\"stdout\",\"data\":\" -------------------------------------\\r\\n Your code has been rated at -20.00/10\\r\\n \\r\\n\"}"
[...]

PR #71, opened 2021-12-10
Unwrap Nomad Error for Allocation Exec

MrSerth:
As #44 is still not fixed, I investigated further and think we need to unwrap the error before inspecting the close cause. Hence, this MR will allow us to inspect whether the WebSocket connection was closed normally by unwrapping the error first.

Sentry:
```
error executing command in allocation: websocket closed before receiving exit code: websocket: close 1000 (normal)
```

We see at least two wrapped errors in Sentry:

Poseidon:
https://github.com/openHPI/poseidon/blob/ebbbfdb9befe98062276f4f815dc6465b7435ea5/internal/nomad/api_querier.go#L109

[Nomad](https://github.com/hashicorp/nomad/blob/e694000f4674139ab5d25607f5c8b16af3e4e0ac/api/allocations_exec.go#L206):
```go
errCh <- fmt.Errorf("websocket closed before receiving exit code: %w", err)
```

The outer one is added after the check, while the inner one is added through Nomad.

### DIFF
--- a/internal/nomad/api_querier.go
+++ b/internal/nomad/api_querier.go
@@ -100,11 +100,14 @@ func (nc *nomadAPIClient) Execute(runnerID string,
 		return 1, fmt.Errorf("error retrieving allocation info: %w", err)
 	}
 	exitCode, err := nc.client.Allocations().Exec(ctx, allocation, TaskName, tty, command, stdin, stdout, stderr, nil, nil)
-	if err != nil && websocket.IsCloseError(err, websocket.CloseNormalClosure) {
-		log.WithError(err).Info("The exit code could not be received.")
-		return 0, nil
-	} else if err != nil {
-		return 1, fmt.Errorf("error executing command in allocation: %w", err)
+	if err != nil {
+		rootCause := errors.Unwrap(err)
+		if rootCause != nil && websocket.IsCloseError(rootCause, websocket.CloseNormalClosure) {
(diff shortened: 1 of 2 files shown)


mpass99:
Thanks for investigating further and fixing this issue!

This behavior of `websocket.IsCloseError` is unexpected for me as it seems to me to be the default in error type checks to always unwrap the passed error itself, as it is also done in the [`errors.Is`](https://pkg.go.dev/errors#Is) function.

To show this confusion we could add an `api_querier_test.go` test:

``` Go
package nomad

import (
	"errors"
	"fmt"
	"github.com/gorilla/websocket"
	"github.com/stretchr/testify/assert"
	"testing"
)

func TestWebsocketErrorNeedsToBeUnwrapped(t *testing.T) {
	rawError := &websocket.CloseError{Code: websocket.CloseNormalClosure}
	err := fmt.Errorf("websocket closed before receiving exit code: %w", rawError)

	assert.False(t, websocket.IsCloseError(err, websocket.CloseNormalClosure))
	rootCause := errors.Unwrap(err)
	assert.True(t, websocket.IsCloseError(rootCause, websocket.CloseNormalClosure))
}

```

MrSerth:
Alright, that's a good addition. I added the test which confirmed our assumption.

mpass99 (approved):

PR #73, opened 2021-12-12
Fix/#59 memory leak

### LINKED ISSUE #59: Potential Memory Leak
We might have a memory leak in Poseidon where memory is not freed correctly. As a consequence, the virtual machine freezer and the service became unavailable.

Memory consumption:
Bildschirmfoto 2021-12-01 um 17 48 05

Deployments (including a restart of the Poseidon service):
Bildschirmfoto 2021-12-01 um 17 48 23

mpass99:
Closes #59 

> We might have a memory leak in Poseidon where memory is not freed correctly.

We leak go routines. Not once per execution, but three times. In this Log [`leakingGoroutines.log`](https://github.com/openHPI/poseidon/files/7698446/leakingGoroutines.log) we see that of three types of goroutines, the count is as large as the number of executions, while the other goroutines take up only a small count.

**Leak 1 in readInputLoop**

https://github.com/openHPI/poseidon/blob/f45d28d4a43f70e1aff7fc9091b780c5a2a61e64/internal/api/websocket.go#L69-L77

When the context in line 74 is done nobody will ever read the `readMessage` and so the goroutine of line 69 is stuck in line 71.

**Leak 2 in the nullio reader**

https://github.com/openHPI/poseidon/blob/f45d28d4a43f70e1aff7fc9091b780c5a2a61e64/pkg/nullio/nullio.go#L11-L14

The nullio Reader is meant to be blocking. But if there is no end at all we create more and more go routines all stucked in line 13.

**Leak 3 in the Read for the executor**

https://github.com/openHPI/poseidon/blob/f45d28d4a43f70e1aff7fc9091b780c5a2a61e64/internal/api/websocket.go#L129-L133

There is some point where nothing needs to be written to the executor. Then this select will block forever.
This fix was a bit more complex, because our proxy should close the connection to the client at a certain point, but a separate go routine still wants to send a SIGQUIT to the Runner after that.

### DIFF
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -61,14 +61,20 @@ func newCodeOceanToRawReader(connection webSocketConnection) *codeOceanToRawRead
 // CloseHandler.
 func (cr *codeOceanToRawReader) readInputLoop(ctx context.Context) {
 	readMessage := make(chan bool)
-	for {
+	readingContext, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	for ctx.Err() == nil {
 		var messageType int
 		var reader io.Reader
 		var err error
 
 		go func() {
 			messageType, reader, err = cr.connection.NextReader()
-			readMessage <- true
+			select {
+			case <-readingContext.Done():
+			case readMessage <- true:
+			}
 		}()
 		select {
 		case <-ctx.Done():
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -30,13 +30,16 @@ type webSocketConnection interface {
 type WebSocketReader interface {
 	io.Reader
 	io.Writer
-	startReadInputLoop() context.CancelFunc
+	startReadInputLoop()
 }
 
 // codeOceanToRawReader is an io.Reader implementation that provides the content of the WebSocket connection
 // to CodeOcean. You have to start the Reader by calling readInputLoop. After that you can use the Read function.
 type codeOceanToRawReader struct {
 	connection webSocketConnection
+
+	// ctx is used to cancel the reading routine.
+	ctx context.Context
 
 	// A buffered channel of bytes is used to store data coming from CodeOcean via WebSocket
 	// and retrieve it when Read(..) is called. Since channels are thread-safe, we use one here
@@ -48,9 +51,10 @@ type codeOceanToRawReader struct {
 	priorityBuffer chan byte
 }
 
-func newCodeOceanToRawReader(connection webSocketConnection) *codeOceanToRawReader {
+func newCodeOceanToRawReader(connection webSocketConnection, ctx context.Context) *codeOceanToRawReader {
 	return &codeOceanToRawReader{
 		connection:     connection,
+		ctx:            ctx,
 		buffer:         make(chan byte, CodeOceanToRawReaderBufferSize),
 		priorityBuffer: make(chan byte, CodeOceanToRawReaderBufferSize),
 	}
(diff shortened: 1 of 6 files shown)


internal/api/websocket.go
```diff
@@ -11,9 +11,15 @@ import (
...
 	"sync"
+	"time"
 )
 
-const CodeOceanToRawReaderBufferSize = 1024
+const (
+	CodeOceanToRawReaderBufferSize = 1024
+	// ProxyShutdownGracePeriod it the period the proxy provides other goroutines to send last messages (like SIGQUIT)
+	// to the Executor.
+	ProxyShutdownGracePeriod = time.Second
```
MrSerth, line 21: That's **one** second, right?

internal/api/websocket.go
```diff
@@ -59,12 +69,14 @@ func newCodeOceanToRawReader(connection webSocketConnection) *codeOceanToRawRead
...
 // This is necessary because input must be read for the connection to handle special messages like close and call the
 // CloseHandler.
-func (cr *codeOceanToRawReader) readInputLoop(ctx context.Context) {
+func (cr *codeOceanToRawReader) readInputLoop() {
 	readMessage := make(chan bool)
-	readingContext, cancel := context.WithCancel(ctx)
+	loopContext, cancel := context.WithCancel(cr.ctx)
+	defer cancel()
+	readingContext, cancel := context.WithCancel(loopContext)
 	defer cancel()
```
MrSerth: Would it make sense to have two different names for these `cancel` actions?

internal/api/websocket.go
```diff
@@ -239,19 +247,18 @@ func newWebSocketProxy(connection webSocketConnection) (*webSocketProxy, error)
...
-		wp.cancelWebSocketWrite()
-		log.Info("Execution returned")
-	case <-wp.userExit:
-		cancelInputLoop()
-		wp.cancelWebSocketWrite()
-		cancelExecution()
+	case <-wp.webSocketCtx.Done():
 		log.Info("Client closed the connection")
+		cancelExecution()
+		<-time.After(ProxyShutdownGracePeriod)
```
MrSerth, line 257: You probably tried: We cannot use a channel to wait until the SIGQUIT was sent, right? Waiting a second here is fine for me as well ;)

pkg/nullio/nullio.go
```diff
@@ -1,16 +1,24 @@
...
 
 func (r Reader) Read(_ []byte) (int, error) {
-	// An empty select blocks forever.
-	select {}
+	if r.Ctx.Err() != nil {
+		return 0, io.EOF
+	}
+
+	<-r.Ctx.Done()
+	return 0, io.EOF
```
MrSerth, line 21: I've read these lines (and also the previous implementation) multiple times and still don't get how reading actually works here. Maybe the intention of the whole Reader in the nullio package is not super clear to me.

pkg/nullio/nullio_test.go
```diff
@@ -1,21 +1,23 @@
...
+	readingContext, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
+	reader := &Reader{readingContext}
 	readerReturned := make(chan bool)
 	go func() {
 		p := make([]byte, 0, 5)
-		_, err := reader.Read(p)
-		require.NoError(t, err)
+		_, _ = reader.Read(p) //nolint:errcheck // We expect an io.EOF error after the readingContext is done.
```
MrSerth, line 20: Should we expect the io.EOF once reading finished?

MrSerth (commented):
Just a few remarks, probably nothing important. I tried by best to understand the context / channels and learned something during the review. But I don't want to state that I would be able to find any important remark or potential issue...

MrSerth:
Nice finding, thanks! I didn't know of the possibilities to analyze the leakage of Go routines with the log you attached. So thanks for pointing that out and explaining the effects of the issues you found.

Please take care to keep the three commits when merging (so that we have them in the main branch and can track the individual changes there). I was also reviewing commit-by-commit.

internal/api/websocket.go
mpass99: Yes, I tried. But I always got lost in too many changes.
Now I have added a better solution.

pkg/nullio/nullio.go
MrSerth: Ah, yes, that makes sense. Should we add your explanation to the code / some documentation? 🙏

internal/api/websocket.go
MrSerth: I like that, it's much cleaner ;)

MrSerth (approved):
I tested the changes locally (without checking the memory consumption) and think they work as expected. Thanks!

PR #76, opened 2021-12-16
Fix early write

### LINKED ISSUE #54: HTTP error before end of send yielding connection reset error in Linux
The request to copy files to a runner might fail even before the content of all files (serialized in a JSON) was sent completely to Poseidon. Depending on the operating system Ruby is running on, this might yield an `Errno::ECONNRESET (Connection reset by peer)` error if the runner was not found. Instead of requesting a new runner, this error is interpreted as a severe exception.

<hr>

A new exercise in the R environment contains a few larger files, including a CSV file with 2.000 lines and a TXT file with 20.000 lines. Encoded in base64, the message is about 750 KB large. As known, the message is used in the HTTP body request to the `{runnerID}/files` route in order to prepare the file system. If a runner is found, everything proceeds as expected and the execution can continue. However, when the runner is not found, the system behaves unexpectedly:

Poseidon returns a 404 error correctly and logs this accordingly. The only problem is that this error is returned from Poseidon before the API client is able to send the whole body. Depending on the HTTP client, this is interpreted as a fatal error and a `Connection reset by peer` exception is thrown. If this happens, the HTTP status code 404 is not interpreted at all.

I spent some time reproducing this issue with Postman, curl and Ruby and with different operating systems:

- Postman on macOS works fine.
- curl on macOS logs an information message:
  `== Info: HTTP error before end of send, stop sending`
- curl on Linux (Ubuntu 20.04 and Kali 2021.3) logs an information message:
  `== Info: HTTP error before end of send, stop sending`
- Ruby (2.7.4) on macOS works fine
- Ruby (2.7.4) on Linux (Ubuntu 20.04 and Kali 2021.3) generates an exception:
  `Errno::ECONNRESET (Connection reset by peer)`

I used the following commands / code snippets:

Curl:
```shell
curl -k --tlsv1.3 -vvv --location --request PATCH 'https://poseidon.hostname.de:7200/api/v1/runners/123e4567-e89b-12d3-a456-426614174000/files' \
--header 'Content-Type: application/json' \
--header 'Poseidon-Token: tokenvalue' \
--data-binary '@/absolute/path/to/data.json' --trace-ascii ./output.log
```

Ruby:
```ruby
require "uri"
require "json"
require "net/http"

url = URI("https://poseidon.hostname.de:7200/api/v1/runners/123e4567-e89b-12d3-a456-426614174000/files")

https = Net::HTTP.new(url.host, url.port)
https.use_ssl = true
https.verify_mode = OpenSSL::SSL::VERIFY_NONE

request = Net::HTTP::Patch.new(url)
request["Content-Type"] = "application/json"
request["Poseidon-Token"] = "tokenvalue"
request.body = File.read("data.json")

response = https.request(request)
puts response.read_body
```

Data: [data.json.zip](https://github.com/openHPI/poseidon/files/7617894/data.json.zip)

<hr>

I don't have a specific clue on the reasoning behind this issue. My best clue is the window size that is exceeded by the large HTTP body. My curl log shows:

```
0000: PATCH /api/v1/runners/123e4567-e89b-12d3-a456-426614174000/files
0040:  HTTP/2
0049: Host: poseidon.hostname.de:7200
0079: user-agent: curl/7.80.0
0092: accept: */*
009f: content-type: application/json
00bf: poseidon-token: tokenvalue
00f1: content-length: 798968
0109: 

[...]
=> Send data, 65535 bytes (0xffff)
[...]
== Info: Connection state changed (MAX_CONCURRENT_STREAMS == 250)!
[...]
=> Send data, 65536 bytes (0x10000)
[...]

// incomple JSON echo in output

0000: HTTP/2 404 
<= Recv header, 32 bytes (0x20)
0000: content-type: application/json
<= Recv header, 20 bytes (0x14)
0000: content-length: 42
<= Recv header, 37 bytes (0x25)
0000: date: Mon, 29 Nov 2021 00:08:32 GMT
== Info: HTTP error before end of send, stop sending
=> Send SSL data, 5 bytes (0x5)
0000: .....
=> Send SSL data, 1 bytes (0x1)
0000: .
<= Recv header, 2 bytes (0x2)
0000: 
<= Recv SSL data, 5 bytes (0x5)
0000: ....Q
<= Recv SSL data, 1 bytes (0x1)
0000: .
<= Recv data, 42 bytes (0x2a)
0000: {"message":"no runner found with this id"}
== Info: Connection #0 to host poseidon.hostname.de left intact
```

As Poseidon is able to return the 404 based on the path (for which it does not require the body), I would assume that we observe a timing issue and different implementations of the "early return" in Linux and macOS.

In Ruby, the following exception occurs:

```irb
2.7.4 :016 > response = https.request(request)
Traceback (most recent call last):
       16: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http.rb:1485:in `block in request'
       15: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http.rb:1492:in `request'
       14: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http.rb:1519:in `transport_request'
       13: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http.rb:1519:in `catch'
       12: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http.rb:1521:in `block in transport_request'
       11: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http/generic_request.rb:123:in `exec'
       10: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/http/generic_request.rb:189:in `send_request_with_body'
        9: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:247:in `write'
        8: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:265:in `writing'
        7: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:248:in `block in write'
        6: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:275:in `write0'
        5: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:275:in `each_with_index'
        4: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:275:in `each'
        3: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/net/protocol.rb:277:in `block in write0'
        2: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/openssl/buffering.rb:383:in `write_nonblock'
        1: from /usr/local/rvm/rubies/ruby-2.7.4/lib/ruby/2.7.0/openssl/buffering.rb:383:in `syswrite_nonblock'
Errno::ECONNRESET (Connection reset by peer)
```

The Faraday library used in Production will log the following error instead: `#<Faraday::ConnectionFailed wrapped=#<Errno::ECONNRESET: Connection reset by peer>>`

mpass99:
Because some clients fail receiving a response before the body is sent entirely. See https://github.com/openHPI/poseidon/issues/54

As discussed, we think that Poseidon is behaving correctly so far. However, in order to easily solve the problem, we simply read the body before we respond.

Closes #54

### DIFF
--- a/internal/api/runners.go
+++ b/internal/api/runners.go
@@ -8,6 +8,7 @@ import (
 	"github.com/openHPI/poseidon/internal/config"
 	"github.com/openHPI/poseidon/internal/runner"
 	"github.com/openHPI/poseidon/pkg/dto"
+	"io"
 	"net/http"
 	"net/url"
 )
@@ -131,6 +132,12 @@ func (r *RunnerController) findRunnerMiddleware(next http.Handler) http.Handler 
 		runnerID := mux.Vars(request)[RunnerIDKey]
 		targetRunner, err := r.manager.Get(runnerID)
 		if err != nil {
+			// We discard the request body because an early write causes errors for some clients.
+			// See https://github.com/openHPI/poseidon/issues/54
+			_, readErr := io.ReadAll(request.Body)
+			if readErr != nil {
+				log.WithError(readErr).Warn("Failed to discard the request body")
+			}
 			writeNotFound(writer, err)
 			return
 		}
--- a/internal/api/runners_test.go
+++ b/internal/api/runners_test.go
@@ -17,6 +17,8 @@ import (
 	"testing"
 )
 
+const invalidID = "some-invalid-runner-id"
+
 type MiddlewareTestSuite struct {
 	suite.Suite
 	manager        *runner.ManagerMock
@@ -67,13 +69,28 @@ func (s *MiddlewareTestSuite) TestFindRunnerMiddlewareIfRunnerExists() {
 }
 
 func (s *MiddlewareTestSuite) TestFindRunnerMiddlewareIfRunnerDoesNotExist() {
-	invalidID := "some-invalid-runner-id"
 	s.manager.On("Get", invalidID).Return(nil, runner.ErrRunnerNotFound)
 
 	recorder := httptest.NewRecorder()
 	s.router.ServeHTTP(recorder, s.runnerRequest(invalidID))
 
 	s.Equal(http.StatusNotFound, recorder.Code)
+}
+
+func (s *MiddlewareTestSuite) TestFindRunnerMiddlewareDoesNotEarlyRespond() {
+	body := strings.NewReader(strings.Repeat("A", 798968))
+
+	path, err := s.router.Get("test-runner-id").URL(RunnerIDKey, invalidID)
+	s.Require().NoError(err)
+	request, err := http.NewRequest(http.MethodPost, path.String(), body)
+	s.Require().NoError(err)
+
+	s.manager.On("Get", mock.AnythingOfType("string")).Return(nil, runner.ErrRunnerNotFound)
+	recorder := httptest.NewRecorder()
+	s.router.ServeHTTP(recorder, request)
+
+	s.Equal(http.StatusNotFound, recorder.Code)
+	s.Equal(0, body.Len()) // No data should be unread
 }
 
 func TestRunnerRouteTestSuite(t *testing.T) {
@@ -261,7 +278,6 @@ func (s *UpdateFileSystemRouteTestSuite) TestUpdateFileSystemReturnsBadRequestOn
 }
 
 func (s *UpdateFileSystemRouteTestSuite) TestUpdateFileSystemToNonExistingRunnerReturnsNotFound() {
-	invalidID := "some-invalid-runner-id"
 	s.runnerManager.On("Get", invalidID).Return(nil, runner.ErrRunnerNotFound)
 	path, err := s.router.Get(UpdateFileSystemPath).URL(RunnerIDKey, invalidID)
 	s.Require().NoError(err)


MrSerth (approved):
Awesome, thank you! This will finally resolve my headache I had with #54 (after too much debugging...)

Generally speaking, I am able to confirm that Poseidon behaves correctly and that an early response _should_ work. It's sad that just Ruby on Linux has problems with this (even though it works on macOS).

Long story short: Nice job. The code looks straight forward and I like that you added a test. I reproduced the original issue in the staging environment and was able to confirm that this fix will resolve it.

PR #77, opened 2021-12-16
Recreate runners on execution environment update

### LINKED ISSUE #48: Updating the environment leads to too many runners
Updating an execution environment consists of three steps.
Register the environment at Nomad, update the existing runners and scale the number of runners if required.

https://github.com/openHPI/poseidon/blob/705d35802707c777e27b324501fe4731416d2937/internal/environment/manager.go#L130-L141

In some cases the runners (/Allocations) need to be restarted in step two. But (even if our request to Nomad is synchronous) the allocations are restarted asynchronously. Poseidon is informed about the shutdown and the restart of the allocations. This time span can also be longer for example when a new docker image needs to be pulled.

But since the requests are finished early we finish step two before all runners are restarted. Then in step three Poseidon thinks it needs more runners and request new ones. This may lead to the pool size being doubled when an environment is updated, even though this was not requested.

### LINKED ISSUE #69: Allow subsequent updates to Runner environment
With #33, we allow setting the `force_pull` option to `true` to update an execution environment. This works fine for the first update but does not work well for the second or any subsequent update (as the `force_pull` option is already true for the second update).

It will work, when the runners were used in-between, as new runners are usually created with the `force_pull: false` option. Then, updating the runners works as expected. However, we still face this issue for the template job which is not changed in between. We might need to continue using the `force_pull` option but also recreate some allocations (e.g. the template job).

mpass99:
Closes #69

Too many runners are created because the existing runners are updated. During the update, there is a short time window in which the runner is not available. Since we cannot wait completely for the update, the scaling process is already started, which, due to the runners just restarting, calculates the number of runners needed incorrectly and creates too many. By removing the update process, the cause of the miscalculation is removed.

Closes #48

We require Nomad to re-fetch the Docker image every time the environment is updated. So far, this is done by changing the jobHCL. However, we only do this once and therefore the new image is no longer loaded with the second update. By restarting the runners with each update of the execution environment, the latest Docker image is always loaded.

### DIFF
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -68,7 +68,7 @@ func (n *NomadEnvironment) SetID(id dto.EnvironmentID) {
 }
 
 func (n *NomadEnvironment) PrewarmingPoolSize() uint {
-	configTaskGroup := nomad.FindOrCreateConfigTaskGroup(n.job)
+	configTaskGroup := nomad.FindAndValidateConfigTaskGroup(n.job)
 	count, err := strconv.Atoi(configTaskGroup.Meta[nomad.ConfigMetaPoolSizeKey])
 	if err != nil {
 		log.WithError(err).Error("Prewarming pool size can not be parsed from Job")
@@ -77,7 +77,7 @@ func (n *NomadEnvironment) PrewarmingPoolSize() uint {
 }
 
 func (n *NomadEnvironment) SetPrewarmingPoolSize(count uint) {
-	taskGroup := nomad.FindOrCreateConfigTaskGroup(n.job)
+	taskGroup := nomad.FindAndValidateConfigTaskGroup(n.job)
 
 	if taskGroup.Meta == nil {
 		taskGroup.Meta = make(map[string]string)
@@ -86,36 +86,36 @@ func (n *NomadEnvironment) SetPrewarmingPoolSize(count uint) {
 }
 
 func (n *NomadEnvironment) CPULimit() uint {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 	return uint(*defaultTask.Resources.CPU)
 }
 
 func (n *NomadEnvironment) SetCPULimit(limit uint) {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 
 	integerCPULimit := int(limit)
 	defaultTask.Resources.CPU = &integerCPULimit
 }
 
 func (n *NomadEnvironment) MemoryLimit() uint {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 	return uint(*defaultTask.Resources.MemoryMB)
 }
 
 func (n *NomadEnvironment) SetMemoryLimit(limit uint) {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 
 	integerMemoryLimit := int(limit)
 	defaultTask.Resources.MemoryMB = &integerMemoryLimit
 }
 
 func (n *NomadEnvironment) Image() string {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 	image, ok := defaultTask.Config["image"].(string)
 	if !ok {
 		image = ""
@@ -124,14 +124,14 @@ func (n *NomadEnvironment) Image() string {
 }
 
 func (n *NomadEnvironment) SetImage(image string) {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 	defaultTask.Config["image"] = image
 }
 
 func (n *NomadEnvironment) NetworkAccess() (allowed bool, ports []uint16) {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 
 	allowed = defaultTask.Config["network_mode"] != "none"
 	if len(defaultTaskGroup.Networks) > 0 {
@@ -144,8 +144,8 @@ func (n *NomadEnvironment) NetworkAccess() (allowed bool, ports []uint16) {
 }
 
 func (n *NomadEnvironment) SetNetworkAccess(allow bool, exposedPorts []uint16) {
-	defaultTaskGroup := nomad.FindOrCreateDefaultTaskGroup(n.job)
-	defaultTask := nomad.FindOrCreateDefaultTask(defaultTaskGroup)
+	defaultTaskGroup := nomad.FindAndValidateDefaultTaskGroup(n.job)
+	defaultTask := nomad.FindAndValidateDefaultTask(defaultTaskGroup)
 
 	if len(defaultTaskGroup.Tasks) == 0 {
 		// This function is only used internally and must be called as last step when configuring the task.
@@ -284,6 +284,9 @@ func (n *NomadEnvironment) DeepCopyJob() *nomadApi.Job {
 	return copyEnvironment.job
 }
 
+// SetConfigFrom gets the options from the environment job and saves it into another temporary job.
+// IMPROVE: The getters use a validation function that theoretically could edit the environment job.
+// But this modification might never been saved to Nomad.
 func (n *NomadEnvironment) SetConfigFrom(environment runner.ExecutionEnvironment) {
 	n.SetID(environment.ID())
 	n.SetPrewarmingPoolSize(environment.PrewarmingPoolSize())
--- a/internal/environment/environment.go
+++ b/internal/environment/environment.go
@@ -3,7 +3,6 @@ package environment
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"github.com/google/uuid"
 	nomadApi "github.com/hashicorp/nomad/api"
@@ -12,14 +11,11 @@ import (
 	"github.com/openHPI/poseidon/internal/runner"
 	"github.com/openHPI/poseidon/pkg/dto"
 	"strconv"
+	"sync"
 )
 
 const (
 	portNumberBase = 10
-)
-
-var (
-	ErrorUpdatingExecutionEnvironment = errors.New("errors occurred when updating environment")
 )
 
 type NomadEnvironment struct {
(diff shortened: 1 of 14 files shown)


internal/environment/environment.go
```diff
@@ -173,6 +169,11 @@ func (n *NomadEnvironment) SetNetworkAccess(allow bool, exposedPorts []uint16) {
 // Register creates a Nomad job based on the default job configuration and the given parameters.
 // It registers the job with Nomad and waits until the registration completes.
 func (n *NomadEnvironment) Register(apiClient nomad.ExecutorAPI) error {
+	// To avoid docker image issues. See https://github.com/openHPI/poseidon/issues/69
+	if err := n.Delete(apiClient); err != nil {
+		return fmt.Errorf("failed to remove the environment: %w", err)
+	}
+
```
MrSerth, line 176: I'm not sure whether this is the correct place. I would argue that removing the environment in a method called "Register" is unexpected. Wouldn't it make sense to put these lines to the manager? 

https://github.com/openHPI/poseidon/blob/dd7b150e55b9b91c6d0b3edae0c5e50fa3f91de1/internal/environment/manager.go#L133

Maybe we could also split the `CreateOrUpdate` method to make the purpose even more clear:

```go
// Fetch or create execution environment
environment, existing_environment := m.runnerManager.GetEnvironment(id)
if existing_environment {
    // Remove existing environment to force downloading the newest Docker image
    // See https://github.com/openHPI/poseidon/issues/69
    err = environment.Delete(m.api)
    if err != nil {
        return fmt.Errorf("failed to remove the environment: %w", err)
     }
}

// Create a new environment with the given ID // could be a dedicated method?
environment, err = NewNomadEnvironment(m.templateEnvironmentHCL)
if err != nil {
    return false, fmt.Errorf("error creating Nomad environment: %w", err)
}
environment.SetID(id)


// Set options according to request // maybe extract to separate method or together with `NewNomadEnvironment`  part?
environment.SetPrewarmingPoolSize(request.PrewarmingPoolSize)
environment.SetCPULimit(request.CPULimit)
environment.SetMemoryLimit(request.MemoryLimit)
environment.SetImage(request.Image)
environment.SetNetworkAccess(request.NetworkAccess, request.ExposedPorts)

// Keep a copy of environment specification in memory // maybe rename method and remove return value (isn't this similar to `!existing_environment `)?
m.runnerManager.StoreEnvironment(environment)

// Register template Job with Nomad
err = environment.Register(m.api)
if err != nil {
    return false, fmt.Errorf("error registering template job in API: %w", err)
}

// Launch idle runners based on template job
err = environment.LaunchIdleRunners(m.api) // maybe rename method?
if err != nil {
    return false, fmt.Errorf("error scaling template job in API: %w", err)
}

return existing_environment, nil
```

Just a proposal to have the important things in one place for creating / updating runners. I didn't test the code, so take it with a grain of salt.

internal/environment/environment.go
```diff
@@ -205,36 +206,10 @@ func (n *NomadEnvironment) Scale(apiClient nomad.ExecutorAPI) error {
 	if required > 0 {
 		return n.createRunners(apiClient, uint(required), true)
 	} else {
-		return n.removeRunners(apiClient, uint(-required))
+		return n.removeIdleRunners(apiClient, uint(-required))
```
MrSerth, line 209: If we always delete the execution environment incl. all runners, do we need to keep the `else` path in the `Scale` method? Is there any use case we could decrease the scale without removing the environment before?

internal/environment/environment.go
```diff
@@ -361,3 +336,29 @@ func (n *NomadEnvironment) removeRunners(apiClient nomad.ExecutorAPI, count uint
 	}
 	return nil
 }
+
+func (n *NomadEnvironment) removeRunners(apiClient nomad.ExecutorAPI) error {
+	// Only to avoid timing issues as an idle runner is also removed when Nomad has deleted the allocation.
+	for _, r := range n.idleRunners.List() {
+		n.idleRunners.Delete(r.ID())
+	}
```
MrSerth, line 344: Could you please add a sentence _what_ these lines are intended to do (in addition to _why_ they are included)? If I got it right, I would state something like:

```suggestion
        // Purge the list of available idle runners for the given execution environment // maybe own method in storage.go?
        // This prevents a race condition where an idle runner is returned while removing the environment
	for _, r := range n.idleRunners.List() {
		n.idleRunners.Delete(r.ID())
	}
```

internal/environment/environment.go
```diff
@@ -361,3 +336,29 @@ func (n *NomadEnvironment) removeRunners(apiClient nomad.ExecutorAPI, count uint
...
+
+	ids, err := apiClient.LoadRunnerIDs(nomad.RunnerJobID(n.ID(), ""))
+	if err != nil {
+		return fmt.Errorf("failed to load runner ids: %w", err)
+	}
+
+	var wg sync.WaitGroup
```
MrSerth, line 351:
```suggestion
        // Remove all (idle and used) runners for the given environment and block execution until Nomad confirmed the deletion request
	var wg sync.WaitGroup
```

internal/environment/environment.go
```diff
@@ -205,36 +206,10 @@ func (n *NomadEnvironment) Scale(apiClient nomad.ExecutorAPI) error {
 	if required > 0 {
 		return n.createRunners(apiClient, uint(required), true)
 	} else {
-		return n.removeRunners(apiClient, uint(-required))
+		return n.removeIdleRunners(apiClient, uint(-required))
 	}
 }
 
-func (n *NomadEnvironment) UpdateRunnerSpecs(apiClient nomad.ExecutorAPI) error {
```
MrSerth, line 212: That method still exists in `execution_environment_mock.go`

MrSerth (commented):
Thanks! I included some feedback from my code review that I generated by reading and /trying to) understand the changes (and the whole environment / runner creation as a whole). My main request would be to check the structure of these changes and to evaluate how we could improve the readability of the code. You'll find some suggestions in my comments - looking forward to your honest feedback 😊

internal/environment/environment.go
mpass99: I have added a description to the new `Purge` function

mpass99 (commented):
Thank you for the comments. I appreciate you drawing my attention to aspects I hadn't considered.

internal/environment/environment.go
mpass99: One reason to leave it as it was would be that several e2e tests would otherwise be flaky and in need of revision.
These tests create the same execution environment several times in a short period of time. Sometimes it happens that a new runner was created from a previous run exactly in the period between the deletion and the scaling. In this case, idle runners have to be removed.
But maybe we should review the e2e tests rather than having unused code running in production!?

internal/environment/environment.go
MrSerth: Ah, thank you! My assumption was not correct but now I get the cause with your extended comment 🙂

internal/nomad/job.go
```diff
@@ -99,7 +99,11 @@ func FindOrCreateConfigTask(taskGroup *nomadApi.TaskGroup) *nomadApi.Task {
 	if task.Config == nil {
 		task.Config = make(map[string]interface{})
 	}
-	task.Config["command"] = ConfigTaskCommand
+	// This function should allow concurrency in the "Find" case.
+	// Therefore, this condition is necessary to remove concurrent writes in the "Find" case.
+	if v, ok := task.Config["command"]; !(ok && v == ConfigTaskCommand) {
+		task.Config["command"] = ConfigTaskCommand
+	}
```
MrSerth, line 106: I am not sure whether I understand these changes. I get that you want to enable concurrency when an existing config job is found and does need to be created. And you want to prevent concurrent writes to the command config of the task.

Now, I am wondering if the code is able to achieve that: What if thread A checks the condition and enters the `if branch` just before line 105 gets executed, another thread B also checks the condition and jumps into line 105. In that case, both threads might still execute line 105, possibly with a concurrent write. Similarly, this could affect the `if task == nil` and `if task.Config == nil` checks. My search for occurrences of concurrent access to these lines was not super successful (without checking the tests but including other callees). Could you please help me out here (and for the similar lines below)?

internal/environment/environment.go
MrSerth: I almost got it and see the issue. Just one question remaining: Who is responsible for these additional (unwanted) runners in the test scenario? Is it caused by an older, asynchronous request from Poseidon to Nomad which gets executed too "late" in Nomad (but has been planned successfully beforehand) or is it Poseidon requesting new runners in that moment? If it is the first case: Why can't we stop these allocations from spawning? If I see that correctly in Nomad, they already have an allocation ID while being pending (which would make them accessible for the delete request). And if it is the second case: Why is Poseidon still requesting these new runners? Is there some concurrent process that is not aware of the deletion?

I wondered whether this bug could also occur in our production environment (based on any of the two assumptions above). Sure, I am also fine with allowing to reduce the Prewarming pool size (instead of returning an error) but still wondered about this behavior.

MrSerth (commented):
Thanks again for taking another look at these changes and refining these. If you ask me, the readability of the code and a few methods increased, especially within the `CreateOrUpdate` method - I hope you agree 😃

internal/environment/environment.go
mpass99: It's your first case. Let's take a look at an example, the e2e test `TestCreateOrUpdateEnvironment`.
- Line 41 the execution environment is created the first time. 
  - Poseidon registers the template job and one additional runner
- Line 50 the execution environment is updated
  - Poseidon updates the template job and registers one additional runner
- Line 59 the execution environment is updated
  - The execution environment is updated
    - It is deleted
    - Now the two runner could be allocated
    - When we now try to scale (and add one additional runner) we should remove one runner (instead)

This behavior can be seen in this log [flakyCase.txt](https://github.com/openHPI/poseidon/files/7739699/flakyCase.txt). The inconspicuous process, on the other hand, generates this log [nonFlakyCase.txt](https://github.com/openHPI/poseidon/files/7739701/nonFlakyCase.txt)

> Why can't we stop these allocations from spawning? If I see that correctly in Nomad, they already have an allocation ID while being pending (which would make them accessible for the delete request).

Just with the logs so far, it is noticeable that the template job for the created environment 42 was never started. By directly telling Nomad to delete the template job (line 211 environment.go), the spawning of the allocation was stopped.
The runners themselves, on the other hand, are only collected via the "LoadRunnerIDs" function. [In a test](https://github.com/openHPI/poseidon/files/7739725/whenIsARunnerIdFetchable.txt), I logged the "Fetched runner ids" until the runners could be listed. It could be determined that this is only the case when (/25ms before) Poseidon knows that the allocation has also been started.

This leads to the two most obvious options, either to cache the IDs of all created runners, or to accept the inaccuracy that when deleting an environment, pending runners are not deleted.

internal/nomad/job.go
mpass99: > What if ...

You are absolutely right!, except we never enter one of the conditions.
These conditions are just guarding clauses but since we have a correct `template-environment-job.hcl` (that includes a config and a default task) and we only edit jobs the correct way, we never use them.

> My search for occurrences of concurrent access to these lines was not super successful

Whenever new runners are requested, we make a copy of the template job. In this process we enter the config (and default) task. In a concurrent way this can happen when two users request a runner in the exact same moment. Noticeable by a slightly flaky behaviour of the `TestTwoSampleAddExactlyTwoRunners` test.

Maybe I shouldn't have smuggled this into this merge request and maybe it would make more sense to validate the job when creating an environment than to do this every time it is accessed.

internal/environment/environment.go
MrSerth: Your log and the specific example is very helpful, thanks! Before going into detail, my main insight is: The tests are not independent of each other and (even worse) there are race conditions between those. I understand where this behavior is coming from but would argue that it is not desirable. I would rather tend to have the tests more independent (for example, in our RSpec tests suite the execution order of tests is randomized for each run forcing each test to be self-contained). While this is completely out of scope for this PR, I think these flaky tests unveiled an issue we should continue to work on.

Back to the example:
I expected that once line 50 gets executed, the previous runner (from line 41) gets deleted anyway so that only one runner is available. Based on your explanation and the log, that runner might not get deleted. It has been scheduled in Nomad but no allocationID is preserved in Poseidon's memory so that it is returned as part of the `LoadRunnerIDs` call (probably because of line 105 in nomad.go?)

To minimize erroneous behavior, I would tend to delete all allocations (even pending one) for the given execution environment. Besides the two options you mentioned (cache IDs, remove runners for outdated execution environments) I thought of adapting the `LoadRunnerIDs` to return all allocation IDs for an environment independent of their status (pending, running, [completed]). Do you think this could work?

internal/nomad/job.go
MrSerth: Okay, I think we're discussing two cases here: The first is about the usage of the guard clause and the second about the concurrent access to the `FindOrCreateConfigTask`. I'll try to address them one by another:

1. Guard Clause
    Okay. The idea is that we only change the value of `task.Config["command"]` if it is not our expecting value of `ConfigTaskCommand`, is it?

2. Concurrent Access
    If I get you right, you argue that the concurrent access might happen if two users request a new runner at the same time (which yes, is a valid use case). In this case, I am wondering if the concurrent write is actually possible: I thought that each runner gets cloned first and any changes are only made to the copy of the template job. This would imply that two threads cannot modify the same object at the same time (because each runner request would first create a copy [read-only access to the template] and then modify the dedicated copy of the template in memory [if required]). And the environment should not be affected by this change, is it?

Probably, the question regarding the concurrent access is the difficult question for me. I am sorry for nagging you here and really tried to have a look on my own but still was unable to find answers to these questions.

internal/environment/environment.go
mpass99: Yes, that is the best option and [it works](https://github.com/openHPI/poseidon/files/7762793/TestCreateOrUpdateEnvironment.log). Thank you :pray:

MrSerth: I am glad it worked out, thank you! 💐

internal/nomad/job.go
MrSerth: Thanks for your comments (and the renaming), they persist the understanding I was able to gather throughout this PR and the discussion we had🦉.

mpass99: 1. Yes
2. The copy was not read only as the getters access the tasks and those access functions (`FindOrCreateConfigTask`) were always overwriting the task config.

As discussed I added some documentation.

MrSerth (approved):
Thanks for the many fixes and changes that are associated with this PR. I didn't thought of the impact and the changes that were "necessary" as part of the PR in the beginning. However, I am very satisfied with the result and the process we took to unveil some potential issues and resolve them as part of the PR review. Thanks again for your assistance in understanding the changes, explaining everything in detail to me and making the adjustments where necessary! 

I also took the chance to validate the changes in our staging environment and didn't discover any issue with the recreation or update of environments. 👍 👍

Issue #78, opened 2021-12-18
Enabled Network fails silently


mpass99:
Currently no execution environments with the options `"networkAccess": true,
    "exposedPorts": [42, 1337]` can be created.
This reveals two errors: The fact that this is not possible (1) and the fact that this happens unnoticed (2).

1. When requesting such an execution environment we can see the failing jobs in the Nomad UI. We can see the error message: `failed to setup alloc: pre-run hook "network" failed: failed to configure networking for alloc: failed to configure network: failed to find plugin "bridge" in path [/opt/cni/bin]`
2. When requesting such an execution environment we get a normal `201 Created` response. When we take a look at [the server log](https://github.com/openHPI/poseidon/files/7739160/cannotCreateExecutionEnvironmentWithEnabledNetwork.txt) we notice that we first receive a `job-register complete` evaluation for the template job and the one requested runner. Then the response to us is written. Only 6s afterwards Poseidon is notified that the allocations actually failed.
Users of Poseidon have no chance to find out about it. Also in the list of Execution Environments, the environment just created is displayed normally. At the most, the Statistic Route could give a hint that the idle runners remain at 0. This way also one e2e test is fooled.

MrSerth:
I added some documentation with #79 and performed these steps on all systems (staging and production). That resolved the immediate issue for now.

PR #79, opened 2021-12-20
Improve documentation for Network support in Nomad

MrSerth:
In order to prevent the issue mentioned in #78, an improved documentation would be helpful. Therefore, this PR adds a documentation on the dependency that needs to be added to all Nomad agents in order to enable network support.

(It's Murphy's law that each PR I open has failing unit tests, e2e tests and Trivy scan results. But why?!)

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -51,3 +51,13 @@ limits {
     http_max_conns_per_client = 0
 }
 ```
+
+### Enable Networking Support in Nomad
+
+In order to allow full networking support in Nomad, the `containernetworking-plugins` are required on the host. They can be either installed manually or through a package manager. In the latter case, the installation path might differ. Hence, add the following line to the `client` directive of the Nomad configuration in `/etc/nomad.d/client.hcl`:
+
+```hcl
+    cni_path = "/usr/lib/cni"
+```
+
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -51,3 +51,13 @@ limits {
...
+
+### Enable Networking Support in Nomad
+
+In order to allow full networking support in Nomad, the `containernetworking-plugins` are required on the host. They can be either installed manually or through a package manager. In the latter case, the installation path might differ. Hence, add the following line to the `client` directive of the Nomad configuration in `/etc/nomad.d/client.hcl`:
+
+```hcl
+    cni_path = "/usr/lib/cni"
+```
+
+If the path is not set up correctly or the dependency is missing, the following error will be shown in Nomad: `failed to find plugin "bridge" in path [/opt/cni/bin]`
```
mpass99, line 63: Thank you for documenting it for future reference.
Another way to document it would be to add the installation of Nomad (and this plugin) to our ansible playbook. This might be addressed in a new issue.

mpass99 (approved):
Thanks, now environments with enabled network can be created again.

docs/configuration.md
MrSerth: That's a good idea, thanks. I created #82 to tackle this somewhere in the future.

Issue #51, opened 2021-11-26
Fix security alerts


mpass99:
We should address our [code scanning issues](https://github.com/openHPI/poseidon/security/code-scanning)

MrSerth:
We should sanitize the log messages according to [this CodeQL rule](https://github.com/github/codeql-go/blob/b2ca04ce1b0b9296b790f621d976f61170ec19ca/ql/src/Security/CWE-117/LogInjection.ql).

PR #81, opened 2021-12-22
Upgrade the dependencies

mpass99:
Unfortunately, the dependency gin-gonic criticised by trivy is still present in version 1.4.0 (and not 1.6.3)

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -8,7 +8,7 @@ require (
 	github.com/gorilla/mux v1.8.0
 	github.com/gorilla/websocket v1.4.2
 	github.com/hashicorp/nomad v1.2.3
-	github.com/hashicorp/nomad/api v0.0.0-20211213195453-35c22bcb6c31
+	github.com/hashicorp/nomad/api v0.0.0-20211221221945-5abe0078ca89
 	github.com/mitchellh/mapstructure v1.4.3
 	github.com/sirupsen/logrus v1.8.1
 	github.com/stretchr/testify v1.7.0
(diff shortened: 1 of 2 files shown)


go.mod
```diff
@@ -40,13 +40,13 @@ require (
 	github.com/hashicorp/go-version v1.3.0 // indirect
 	github.com/hashicorp/golang-lru v0.5.4 // indirect
 	github.com/hashicorp/hcl v1.0.1-vault-3 // indirect
-	github.com/hashicorp/hcl/v2 v2.9.2-0.20210407182552-eb14f8319bdc // indirect
+	github.com/hashicorp/hcl/v2 v2.11.1 // indirect
```
MrSerth, line 43: Without updating this dependency, the compile step should succeed as normal. I noticed that during my previous updates and didn't investigate yet. Otherwise, it looks good; thanks for maintaining the dependencies

MrSerth (commented):
I think we should consider #51 to tackle these issues. So far, I am not too much convinced of Trivy, to be honest, and would evaluate whether we don't understand it or should change it's configuration (or whether we need it at all).

go.mod
mpass99: Thanks, I haven't looked at it yet either....

MrSerth (approved):
Thanks, looks good 😊

go.mod
MrSerth: Fine for me, it might be something we can investigate when resolving the issue with #51. Many (if not all) of our security alerts are related to dependencies and our dependency management.

PR #83, opened 2021-12-22
Add a warning when allocations fail

### LINKED ISSUE #78: Enabled Network fails silently

mpass99:
I have added the filter for `TriggeredBy` so that not every execution produces a log statement ;) Now every failed allocation should produce two log statements as Nomad retries to place the allocation.

Closes #78

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -238,6 +238,10 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 			case <-time.After(resultChannelWriteTimeout):
 				log.WithField("eval", eval).Error("Full evaluation channel")
 			}
+		} else if eval.TriggeredBy == "alloc-failure" {
+			log.WithField("job", eval.JobID).
+				WithField("eval", eval).
+				Warn("Allocation failure")
 		}
 	}
 	return nil
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -238,10 +238,6 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 			case <-time.After(resultChannelWriteTimeout):
 				log.WithField("eval", eval).Error("Full evaluation channel")
 			}
-		} else if eval.TriggeredBy == "alloc-failure" {
-			log.WithField("job", eval.JobID).
-				WithField("eval", eval).
-				Warn("Allocation failure")
 		}
 	}
 	return nil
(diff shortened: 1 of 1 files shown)


mpass99:
There are also alternative solutions to this problem:
We can see in [this log](https://github.com/openHPI/poseidon/files/7763793/eventsDuringAllocationFailure.txt) the status update of the placement failure from Nomad on job, evaluation and allocation level. Therefore we also could:
- Listen to events telling that some job is done and request additional information to decide if the job has failed.
- Listen to both evaluation and job events. Temporarily save if an allocation fails and check when a job is done if its allocation has failed before.
- Check only the evaluation events as we do now but additionally filter for the length of `QueuedAllocations` as we can reduce the number of log statements per failed allocation to / by one (since both attempts seem to differ in that attribute).

MrSerth:
Thanks for the introduction of the log statement and your thoughts to further enhance this issue tracking. 

I reproduced the issue with CodeOcean (even though that required some temporary adaption to Ruby code) and created POSEIDON-11 in Sentry. That also confirms that we log each allocation twice (and maybe for each runner of the Prewarming pool size). I also used your log to understand which events are received by Poseidon :)

Generally speaking, your ideas are worth considering for further debugging and might improve our logging. For this issue here, I would like to keep it simple and maybe add another filter option (as you mentioned in your third option). 

What if we check that the value of the `PreviousEval`? In my test with POSEIDON-11, [the first allocation](https://sentry.io/organizations/openhpi/issues/2878505124/events/oldest/?project=6077585) failed without any `PreviousEval`, while a [retry](https://sentry.io/organizations/openhpi/issues/2878505124/events/c10c63c514ab41efb9eb8b5d5abb0425/?project=6077585) included the allocation of the first attempt as a value of the `PreviousEval`. I'll make a suggestion to the code 👨‍💻

internal/nomad/nomad.go
```diff
@@ -238,6 +238,10 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 			case <-time.After(resultChannelWriteTimeout):
 				log.WithField("eval", eval).Error("Full evaluation channel")
 			}
+		} else if eval.TriggeredBy == "alloc-failure" {
```
MrSerth, line 241:
```suggestion
		} else if eval.TriggeredBy == "alloc-failure" && eval.PreviousEval == "" {
		// Log allocation failures once and not for any retries 
```

mpass99:
What bothered me about the first evaluation-level solution is that we use the magic string `alloc-failure` and our added value of the Sentry message is limited to knowing there was an error - but we have no useful information to determine the cause. For example, in problem case #78, it would not have been logged that it was a network problem.

That's why I went to the allocation level. With the inspiration of your distinguishing feature finding (thanks), I now [systematically](http://www.jsondiff.com/) compared the allocation events and was able to filter the events down to just one using the attributes `FollowupEvalID` and `PreviousAllocation`. Thus we have information for the cause of the error directly in the log / sentry. What do you think?

MrSerth (approved):
That's a great improvement and a nice refactoring! I like it pretty much and think that the enhanced information (with the cause) will be pretty handy for debugging. Thanks for investing the time to continuously working on this 👏

PR #91, opened 2022-01-21
AWS SAM deployment

mpass99:
Add AWS SAM application template. With the sam-cli one can automatically deploy our AWS functions.

ToDo:
- Your AWS Credentials :heavy_check_mark: 
- Authentication (API Key) :man_shrugging:


deploy/aws/events/event.json
```diff
@@ -0,0 +1,22 @@
...
+    "extendedRequestId": "L1Z1CH3rliAFRhg=",
+    "messageDirection": "IN",
+    "messageId": "MUBfpelRFiACF9g=",
+    "requestTime": "12/Jan/2022:13:24:22 +0000",
+    "requestTimeEpoch": 1641993862426,
+    "routeKey": "java11Exec"
+  },
+  "body": "{\n    \"action\": \"java11Exec\",\n    \"cmd\": [\n        \"sh\",\n        \"-c\",\n        \"javac org/example/RecursiveMath.java && java org/example/RecursiveMath\"\n    ],\n    \"files\": {\n        \"org/example/RecursiveMath.java\": \"cGFja2FnZSBvcmcuZXhhbXBsZTsKCnB1YmxpYyBjbGFzcyBSZWN1cnNpdmVNYXRoIHsKCiAgICBwdWJsaWMgc3RhdGljIHZvaWQgbWFpbihTdHJpbmdbXSBhcmdzKSB7CiAgICAgICAgU3lzdGVtLm91dC5wcmludGxuKCJNZWluIFRleHQiKTsKICAgIH0KCiAgICBwdWJsaWMgc3RhdGljIGRvdWJsZSBwb3dlcihpbnQgYmFzZSwgaW50IGV4cG9uZW50KSB7CiAgICAgICAgcmV0dXJuIDQyOwogICAgfQp9Cgo=\"\n    }\n}",
+  "isBase64Encoded": false
+}
```
mpass99, line 22:
```suggestion
}

```

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+    WebSocketMessageType(String name) {
+        this.typeName = name;
+    }
+
+    public String toString() {
+        return typeName;
+    }
+}
+
+// WebSocketMessage is the object that is beeing send over the WebSocket connection.
```
MrSerth, line 45:
```suggestion
// WebSocketMessage is the object that is beeing sent over the WebSocket connection.
```

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+import com.amazonaws.services.lambda.runtime.events.APIGatewayV2WebSocketEvent;
+import com.google.gson.Gson;
+
+// AwsFunctionRequest contains the java files that needs to be executed.
+class AwsFunctionRequest {
+    String[] cmd;
+    Map<String, String> files;
+}
+
+// WebSocketMessageType are the types of messages that are being send back over the WebSocket connection.
```
MrSerth, line 27:
```suggestion
// WebSocketMessageType are the types of messages that are being sent back over the WebSocket connection.
```

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+
+    public static final String disableOutputHeaderKey = "disableOutput";
+
+    // disableOutput: If set to true, no output will be sent over the WebSocket connection.
+    private boolean disableOutput = false;
+
+    public APIGatewayProxyResponseEvent handleRequest(final APIGatewayV2WebSocketEvent input, final Context context) {
+        APIGatewayV2WebSocketEvent.RequestContext ctx = input.getRequestContext();
+        this.gwClient = AmazonApiGatewayManagementApiClientBuilder.standard()
+                .withEndpointConfiguration(new AwsClientBuilder.EndpointConfiguration("https://" + ctx.getDomainName() + "/" + ctx.getStage(), Regions.EU_CENTRAL_1.getName()))
```
MrSerth, line 75: We might want to add a comment somewhere (in the documentation) that EU Central is hard-coded here.

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+            File workingDirectory = this.writeFS(execution.files);
+
+            ProcessBuilder pb = new ProcessBuilder(execution.cmd).redirectErrorStream(true);
+            pb.directory(workingDirectory);
+            Process p = pb.start();
+            InputStream stdout = p.getInputStream(), stderr = p.getErrorStream();
+            this.forwardOutput(p, stdout, stderr);
+            p.destroy();
+            return new APIGatewayProxyResponseEvent().withStatusCode(200);
+        } catch (Exception e) {
```
MrSerth, line 91: Catching any Exception here is very generic but probably fine to "forward" it to the WebSocket consumer

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+    }
+
+    // writeFS writes the files to the local filesystem.
+    private File writeFS(Map<String, String> files) throws IOException {
+        File workspace = Files.createTempDirectory("workspace").toFile();
+        for (Map.Entry<String, String> entry : files.entrySet()) {
+            File f = new File(workspace, entry.getKey());
+
+            boolean ok = f.getParentFile().mkdirs();
+            if (!ok) throw new IOException("Cannot create parent directories.");
```
MrSerth, line 104: Probably not an issue here, but `mkdirs()` will also return false if the directories already existed and thus were not generated. As we expect to have a clean environment, this case should rather happen. Still, it might be worth noticing.

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+        }
+        return workspace;
+    }
+
+    // forwardOutput sends the output of the process to the WebSocket connection.
+    private void forwardOutput(Process p, InputStream stdout, InputStream stderr) throws InterruptedException, IOException {
+        while (p.isAlive() || stdout.available()>0 || stderr.available()>0) {
+            while(stdout.available()>0) this.sendMessage(stdout.readAllBytes(), WebSocketMessageType.WebSocketOutputStdout, 0);
+            while(stderr.available()>0) this.sendMessage(stderr.readAllBytes(), WebSocketMessageType.WebSocketOutputStderr, 0);
+            Thread.sleep(50);
```
MrSerth, line 119: Hm, why are we putting the thread to sleep here? Is that due to prevent busy waiting or some race condition fighting?

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -0,0 +1,136 @@
...
+        this.sendMessage(new byte[0], WebSocketMessageType.WebSocketExit, p.exitValue());
+    }
+
+    // sendMessage sends WebSocketMessage objects back to the requester of this Lambda function.
+    private void sendMessage(byte[] data, WebSocketMessageType type, int exitCode) {
+        if (this.disableOutput) return;
+        WebSocketMessage msg = new WebSocketMessage();
+        msg.data = new String(data);
+        msg.type = type;
+        msg.exitCode = exitCode;
```
MrSerth, line 130: Does that mean every message contains a (faked) exit code even if does not represent a process termination but is part of an output feed? 

And should those message adhere to Poseidon's web socket scheme? This one does not specify the `exitCode` property and simultaneously sets `additionalProperties` to `false`.

deploy/aws/template.yaml
```diff
@@ -0,0 +1,108 @@
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: >
+  PoseidonExecutors
+
+  Execute untrusted code in a AWS functions.
```
MrSerth, line 6:
```suggestion
  Execute untrusted code in AWS functions.
```

deploy/aws/template.yaml
```diff
@@ -0,0 +1,108 @@
...
+    Properties:
+      CodeUri: java11Exec/
+      Handler: poseidon.App::handleRequest
+      Runtime: java11
+      Architectures:
+        - arm64
+      MemorySize: 2048
+      Environment: # More info about Env Vars: https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#environment-object
+        Variables:
+          PARAM1: VALUE
```
MrSerth, line 69: Where is this variable used?

deploy/aws/java11Exec/pom.xml
```diff
@@ -0,0 +1,62 @@
...
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/maven-v4_0_0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>poseidon</groupId>
+    <artifactId>java11Exec</artifactId>
+    <version>1.0</version>
+    <packaging>jar</packaging>
+    <name>A Java executor created for openHPI/Poseidon.</name>
+    <properties>
+        <maven.compiler.source>11</maven.compiler.source>
+        <maven.compiler.target>11</maven.compiler.target>
```
MrSerth, line 11: If I got the [AWS documentation](https://docs.aws.amazon.com/lambda/latest/dg/lambda-java.html) right, Java 11 is the newest version supported by AWS lambda? That is fine and will work for the all Java courses except the newest java-algorithmen2021 course. In the latest course, we switched to the latest JDK available (17.0.1 at that time) and also used language features not available with Java < 16.

MrSerth (commented):
Thanks for preparing the AWS SAM deployment. I found it interesting to have a read at SAM and the code we need to implement for it to work accordingly.

deploy/aws/java11Exec/src/main/java/poseidon/App.java
mpass99: This should prevent busy waiting. Now we use the Scanner class to so.
(Subclassing the InputStream has no effect since the Process create it's own implementation of the InputStream.)

deploy/aws/java11Exec/src/main/java/poseidon/App.java
mpass99: I have oriented myself at the dto.WebSocketMessage but the WebSocket Schema is the right reference point. Thanks!

deploy/aws/template.yaml
mpass99: Just an example value

deploy/aws/README.md
```diff
@@ -0,0 +1,44 @@
+# Poseidon AWS Executors
+
+This project contains source code and supporting files for a serverless application that you can deploy with the SAM CLI. It includes the following functions.
+
+- java11ExecFunction - Code for the application's Lambda function. It can execute Java files.
```
MrSerth, line 5:
```suggestion
- java11ExecFunction - Code for the application's Lambda function. It can execute Java files with JDK 11.
```

deploy/aws/java11Exec/pom.xml
MrSerth: You're welcome. I also made a minor suggestion to add this finding to the documentation.

deploy/aws/java11Exec/src/main/java/poseidon/App.java
MrSerth: I am fine with both. If you have a quick and easy solution to parse the region from the URL or any other parameter, I would do that. If you spend more than 15 / 30 minutes on that, I would add it to the ReadMe.

MrSerth (approved):
LGTM 🏄‍♂️

Issue #43, opened 2021-11-17
Improve Nomad's memory consumption


MrSerth:
When observing Nomad's memory consumption, I noticed the following:

- [x] Allocations reserve as much memory as their limit is set to. In many cases, the whole memory is not required but rather an upper limit for the execution. Hence, we might consider setting a lower memory setting and rather allow [memory overconsumption in Nomad](https://www.nomadproject.io/docs/job-specification/resources#memory-oversubscription).
- [ ] ~~Template jobs consume as much memory as each runner, even when the pool size is set to zero.~~

We should investigate and consider changes to some of the observations made above. This will decrease the environmental impact of Poseidon 🌳. As a consequence, we will be able to serve more users with the same resources.

MrSerth:
This ticket is rather getting important. As we might have up to five (5) parallel courses end of March / beginning of April, this setting will help us to handle more parallel users.

mpass99:
While working on this ticket, I noticed a big drop in memory usage when increasing the pool size. We are still stopping all runners even if we just want more runners.
Did we agree on this last time with the background that environments are not updated so often?

mpass99:
> Template jobs consume as much memory as each runner, even when the pool size is set to zero.

Should we continue to address this where the memory reservation has dropped to 16MB and the CPU reservation to the set level?
If so, we would need to set either the command or the resource consumption unequal to the runners value. We would then store this information in the meta information of the Nomad environment job to be able to recover the environment.

mpass99:
I have deployed Poseidon and the new Nomad configuration in the staging environment.

PR #102, opened 2022-03-11
Enable memory oversubscription

mpass99:
We have to update the Nomad configuration file (staging / prod)

Fixes second task of #43


mpass99:
It work's as expected. For scheduling Nomad uses the config value (16MB) but inside the container you can use up to the environment specific Limit before the execution is killed.

MrSerth (approved):
Awesome, this really looks great! Looking forward to see it live for our courses 🎸

MrSerth:
Closes #43

PR #103, opened 2022-03-12
Add limited support for Makefile parsing.

mpass99:
As the AWS Linux images do not contain make.

### DIFF
--- a/deploy/aws/java11Exec/src/main/java/poseidon/App.java
+++ b/deploy/aws/java11Exec/src/main/java/poseidon/App.java
@@ -5,11 +5,13 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.nio.file.Files;
 import java.util.Base64;
+import java.util.HashMap;
 import java.util.Map;
 import java.util.Scanner;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import com.amazonaws.client.builder.AwsClientBuilder;
-import com.amazonaws.regions.Regions;
 import com.amazonaws.services.apigatewaymanagementapi.AmazonApiGatewayManagementApi;
 import com.amazonaws.services.apigatewaymanagementapi.AmazonApiGatewayManagementApiClientBuilder;
 import com.amazonaws.services.apigatewaymanagementapi.model.PostToConnectionRequest;
@@ -78,7 +80,7 @@ public class App implements RequestHandler<APIGatewayV2WebSocketEvent, APIGatewa
         try {
             File workingDirectory = this.writeFS(execution.files);
 
-            ProcessBuilder pb = new ProcessBuilder(execution.cmd);
+            ProcessBuilder pb = new ProcessBuilder(prepareCommand(execution));
             pb.directory(workingDirectory);
             Process p = pb.start();
             InputStream stdout = p.getInputStream(), stderr = p.getErrorStream();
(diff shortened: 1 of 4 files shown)


deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -120,6 +122,53 @@ private File writeFS(Map<String, String> files) throws IOException {
...
+    private String[] prepareCommand(AwsFunctionRequest execution) {
+        String[] cmd = execution.cmd;
+        String innerCommand = cmd[cmd.length - 1]; // We expect a "sh -c" wrapped command.
+        String replacedInnerCommand = simpleMakefileReplacement(innerCommand, execution.files);
+        cmd[cmd.length - 1] = replacedInnerCommand;
+        return cmd;
+    }
+
+    protected String simpleMakefileReplacement(String command, Map<String, String> files) {
+        Pattern isMakeCommand = Pattern.compile("^make\\s?(\\w*)?$");
```
MrSerth, line 134: We might want to introduce a non-capturing group and (maybe) a named capturing group: `^make(?:\s+(?'name'\w*)?)$#`

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -78,7 +80,7 @@ public APIGatewayProxyResponseEvent handleRequest(final APIGatewayV2WebSocketEve
         try {
             File workingDirectory = this.writeFS(execution.files);
 
-            ProcessBuilder pb = new ProcessBuilder(execution.cmd);
+            ProcessBuilder pb = new ProcessBuilder(prepareCommand(execution));
```
MrSerth, line 83: As discussed, we could investigate whether a Makefile object is suitable here.

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -120,6 +122,53 @@ private File writeFS(Map<String, String> files) throws IOException {
...
+        }
+        String startRule = makeCommand.group(1);
+
+        String makefileB64 = files.get("Makefile");
+        String makefile = new String(Base64.getDecoder().decode(makefileB64), StandardCharsets.UTF_8);
+        Pattern makeRules = Pattern.compile("(.*):\\n((\\t.*\\n)*)");
+        Matcher makeRuleMatcher = makeRules.matcher(makefile);
+
+        Map<String, String[]> rules = new HashMap<>();
+        String firstRule = null;
```
MrSerth, line 148: This could be an object attribute (maybe `private`?)

deploy/aws/java11Exec/src/main/java/poseidon/App.java
```diff
@@ -120,6 +122,53 @@ private File writeFS(Map<String, String> files) throws IOException {
...
+        Matcher makeCommand = isMakeCommand.matcher(command);
+
+        if (!makeCommand.find() || !(files.containsKey("Makefile") || files.containsKey("makefile"))) {
+            return command;
+        }
+        String startRule = makeCommand.group(1);
+
+        String makefileB64 = files.get("Makefile");
+        String makefile = new String(Base64.getDecoder().decode(makefileB64), StandardCharsets.UTF_8);
+        Pattern makeRules = Pattern.compile("(.*):\\n((\\t.*\\n)*)");
```
MrSerth, line 144: The RegEx patterns might be `private static final`

MrSerth (commented):
I like the functionality of the Makefile parsing, let's evaluate whether a Makefile class / object would be suitable as well ;)

deploy/aws/java11Exec/src/main/java/poseidon/App.java
mpass99: Nice regex feature and fancy that you know about them.
How about `^make(?:\s+(?<startRule>\w*))?$` and `(?<name>.*):\n(?<commands>(?:\t.+\n?)*)`

MrSerth: Yes, that looks fine!

deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java
```diff
@@ -0,0 +1,101 @@
...
+            String ruleName = makeRuleMatcher.group("name");
+            if (startRule == null) {
+                startRule = ruleName;
+            }
+
+            String[] ruleCommands = makeRuleMatcher.group("commands").split("\n");
+            String[] trimmedCommands = new String[ruleCommands.length];
+            for (int i = 0; i < ruleCommands.length; i++) {
+                trimmedCommands[i] = ruleCommands[i].trim();
+            }
```
MrSerth, line 86: This is probably equivalent to:

```java
String[] trimmedCommands = Arrays.stream(ruleCommands).map(String::trim).toArray(String[]::new);
```

MrSerth (commented):
Thanks for changing the implementation and creating a Makefile class. I really like the improved readability and the different methods you created to increase readability and maintainability. 

Besides making a small suggestion to use a map function for the trimming of commands, I was thinking of retrieving the commands from the Makefile (with `getCommand()`). In general, I see two options:

1. Having a specific constructor for a make target and the Makefile (two parameters) and one method to get the command (no parameters) as you did. This requires a new Makefile object to get created when another command should get retrieved.
2. A generic constructor for a Makefil (one parameter for the file content) and a method to get specific commands for one rule (`getCommand("run")` vs `getCommand("test")`; that could be combined with the `unwrapCommand` or include that to allow `getCommand(["sh", "-c", "make run"])` maybe with method overloading).

I am not sure about the pros and cons of both methods. The first one is already done and probably serves all use cases that we have in mind today. While I expect it to be slower for several `make` calls with different targets, it is probably faster for retrieving the same commands for a given rule multiple times. For the second solution, these probably swap (better performance when retrieving multiple different targets, as the Makefile is only parsed and prepared once). 

What do you think? I don't necessarily require any change, just wanted to share my thoughts with you.

MrSerth:
Here is a quick (and untested) draft of idea No. 2:

```java

public String getCommand() {
  return getCommand(startRule);
}

public String getCommand(String rule) {
  // Check for specific rule or default to first rule
  if (rule == null) {
    rule = startRule;
  } 
  
  String[] commands = rules.get(rule);
  return wrapCommand(commands);
}

public String getCommand(String[] shellCommand) {
  String rule = unwrapCommand(shellCommand);
  Matcher makeCommandMatcher = isMakeCommand.matcher(makeCommand);
        
  // No make command given
  if (!makeCommandMatcher.find()) {
    throw new NoMakeCommandException();
  }

  // Identify make target
  String ruleArgument = makeCommandMatcher.group("startRule");

  // Get Rules for target
  return getCommand(ruleArgument);
}
```

mpass99:
Nice idea! I have added it with a few minor changes (wrap and unwrap does not belong to the makefile; getCommand() is not used; the remaining two methods need to be distinguished by it's name).

MrSerth (approved):
Great, thanks for taking care and finalizing the rough sketch we created! This looks ready 🏁

PR #108, opened 2022-03-21
Upgrade Go from 1.17 to 1.18

MrSerth:
I just prepared an upgrade from Go 1.17 to Go 1.18 without taking advantage of the new features (that could be a follow-up step).


MrSerth:
Besides Trivy (which I would leave to #51), this seems to look fine.

mpass99 (approved):

PR #109, opened 2022-04-03
Add aws environments to the statistics

### LINKED ISSUE #96: AWS Environment Manager Statistics route
https://github.com/openHPI/poseidon/pull/87#discussion_r803161220

> Of course, listing a prewarming pool size doesn't make much sense and there is no concept of "reserved" runners. We might want to list that "limitation" somewhere because it hides the usage of AWS more to CodeOcean than the usage of Nomad does.

> One could at least return the number of UsedRunners and document this "limitation" of the transparent prewarming.

mpass99:
Closes #96 

Only the field `usedRunners` contains (dynamic) information.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -106,7 +106,7 @@ jobs:
 
   e2e-test:
     runs-on: ubuntu-latest
-    needs: [ compile, dep-scan, test ]
+    needs: [ compile, test ]
     env:
       POSEIDON_AWS_ENABLED: true
       POSEIDON_AWS_ENDPOINT: ${{ secrets.POSEIDON_AWS_ENDPOINT }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,6 +52,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2.5.2
         with:
           version: latest
+          args: --timeout=3m
 
   test:
     runs-on: ubuntu-latest
--- a/internal/environment/aws_environment.go
+++ b/internal/environment/aws_environment.go
@@ -65,8 +65,9 @@ func (a *AWSEnvironment) IdleRunnerCount() int {
 }
 
 // PrewarmingPoolSize is neither supported nor required. It is handled transparently by AWS.
+// For easy compatibility with CodeOcean, 1 is the static value.
 func (a *AWSEnvironment) PrewarmingPoolSize() uint {
-	return 0
+	return 1
 }
 
 // SetPrewarmingPoolSize is neither supported nor required. It is handled transparently by AWS.
--- a/internal/environment/aws_environment.go
+++ b/internal/environment/aws_environment.go
@@ -59,6 +59,11 @@ func (a *AWSEnvironment) Sample() (r runner.Runner, ok bool) {
 
 // The following methods are not supported at this moment.
 
+// IdleRunnerCount is not supported as we have no information about the AWS managed prewarming pool.
+func (a *AWSEnvironment) IdleRunnerCount() int {
+	return 0
+}
+
 // PrewarmingPoolSize is neither supported nor required. It is handled transparently by AWS.
 func (a *AWSEnvironment) PrewarmingPoolSize() uint {
 	return 0
@@ -109,9 +114,5 @@ func (a *AWSEnvironment) AddRunner(_ runner.Runner) {
 }
 
 func (a *AWSEnvironment) DeleteRunner(_ string) {
-	panic("not supported")
-}
-
-func (a *AWSEnvironment) IdleRunnerCount() int {
 	panic("not supported")
 }
--- a/internal/runner/abstract_manager.go
+++ b/internal/runner/abstract_manager.go
@@ -64,7 +64,7 @@ func (n *AbstractManager) EnvironmentStatistics() map[dto.EnvironmentID]*dto.Sta
 			ID:                 int(e.ID()),
 			PrewarmingPoolSize: e.PrewarmingPoolSize(),
 			IdleRunners:        uint(e.IdleRunnerCount()),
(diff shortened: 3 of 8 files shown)


.github/workflows/ci.yml
```diff
@@ -52,6 +52,7 @@ jobs:
         uses: golangci/golangci-lint-action@v2.5.2
```
mpass99, line 52: Why do we prefer the GitHub action over the Makefile command?
Contra:
- You would not assume that the CI does something else than you do.
- Information duplication. The `timeout` flag in the Makefile was probably intended for execution in the CI, but was not used there.

MrSerth: Good question, I don't know. A potential advantage could be to use an action with a dedicated formatter that annotates results in Pull Requests. As we don't use this feature so far, I am also fine with another mechanism.

internal/runner/abstract_manager.go
```diff
@@ -58,7 +58,17 @@ func (n *AbstractManager) DeleteEnvironment(id dto.EnvironmentID) {
...
 
 func (n *AbstractManager) EnvironmentStatistics() map[dto.EnvironmentID]*dto.StatisticalExecutionEnvironmentData {
-	return map[dto.EnvironmentID]*dto.StatisticalExecutionEnvironmentData{}
+	environments := make(map[dto.EnvironmentID]*dto.StatisticalExecutionEnvironmentData)
+	for _, e := range n.environments.List() {
+		environments[e.ID()] = &dto.StatisticalExecutionEnvironmentData{
+			ID:                 int(e.ID()),
+			PrewarmingPoolSize: e.PrewarmingPoolSize(),
+			IdleRunners:        uint(e.IdleRunnerCount()),
+			UsedRunners:        0,
```
MrSerth, line 67: I noticed that these lines were moved from the Nomad manager, but why are we returning 0 for the `UsedRunenrs` here? These are added later as the information is present in the specific runner manager?

MrSerth (approved):
These changes look fine so far, but I haven't executed those. 

Is the CodeOcean dashboard working as expected or does it require further changes? I was wondering because, for AWS, we now have a pool size of 0 and 0 idle runners. Within the dashboard, we provide a visual progress bar where a progress bar indicates the percentage of idle runners in relation to the pool size. I just want to ensure we're not doing a division by zero there (0 idle runners / 0 pool size = 100% filled?).

.github/workflows/ci.yml
mpass99: https://github.com/openHPI/poseidon/issues/111

internal/runner/abstract_manager.go
mpass99: Exactly! Would you like a comment on this?

mpass99:
> I just want to ensure we're not doing a division by zero there

We do, but it doesn't seem to bother Javascript (thanks to the concept NaN?). The progress bar is suitably empty.

https://github.com/openHPI/codeocean/blob/298761885c9a86aef44b40352421bd40aafedf37/app/assets/javascripts/dashboard.js#L87

> Is the CodeOcean dashboard working as expected or does it require further changes?

Unfortunately, the AWS environments are not displayed in the graph, as the prewarming pool size would have to be greater than 0.
It seems to me that an adjustment on both CodeOcean and Poseidon side would not be so nice, but the feature is worth it enough that we make the adjustment on one side?!

https://github.com/openHPI/codeocean/blob/298761885c9a86aef44b40352421bd40aafedf37/app/assets/javascripts/dashboard.js#L71

internal/environment/aws_environment.go
```diff
@@ -59,6 +59,11 @@ func (a *AWSEnvironment) Sample() (r runner.Runner, ok bool) {
...
 // The following methods are not supported at this moment.
 
+// IdleRunnerCount is not supported as we have no information about the AWS managed prewarming pool.
+func (a *AWSEnvironment) IdleRunnerCount() int {
+	return 0
+}
+
 // PrewarmingPoolSize is neither supported nor required. It is handled transparently by AWS.
 func (a *AWSEnvironment) PrewarmingPoolSize() uint {
 	return 0
```
mpass99, line 69:
```suggestion
// For easy compatibility with CodeOcean, 1 is the static value.
func (a *AWSEnvironment) PrewarmingPoolSize() uint {
	return 1
```

internal/runner/abstract_manager.go
MrSerth: Yes, that would be great but think it is optional 👒

MrSerth:
> > I just want to ensure we're not doing a division by zero there
> 
> We do, but it doesn't seem to bother Javascript (thanks to the concept NaN?). The progress bar is suitably empty.
> 
> https://github.com/openHPI/codeocean/blob/298761885c9a86aef44b40352421bd40aafedf37/app/assets/javascripts/dashboard.js#L87

Hm, yes, there is no exception but `NaN` is returned:

Bildschirmfoto 2022-04-08 um 14 50 45

This is sub-optimal, but fine. However, we use this value for the [CSS width attribute](https://github.com/openHPI/codeocean/blob/298761885c9a86aef44b40352421bd40aafedf37/app/assets/javascripts/dashboard.js#L91), where specifying `width: NaN%` doesn't make sense and might allow browsers to display something somehow. Maybe we should prevent that...

internal/runner/abstract_manager.go
mpass99:
```suggestion
			UsedRunners:        0, // Adjusted later as the information is present in the specific runner manager.
```

mpass99:
The Trivy vulnerability scanner keeps crashing when downloading it's database. https://github.com/aquasecurity/trivy/issues/288 Let's give this issue some time and observe if it is on our side or if it was an trivy issue.. #112

Issue #110, opened 2022-04-05
Create Grafana Dashboards


MrSerth:
As a follow-up to #67, we want to create Grafana statistics based on the data we collected. These dashboards should visualize the data and provide a quick overview of:
- a histogram of incoming requests
- status of a request (success / failure / HTTP code)
- insights on the usage of runners (creation, deletion, reusing)

mpass99:
@MrSerth, I would appreciate it if you could set up the new data source in Grafana or give me the necessary permissions..

mpass99:
@MrSerth, I would appreciate it if you could set up the new data source in Grafana or give me the necessary permissions..

MrSerth:
> @MrSerth, I would appreciate it if you could set up the new data source in Grafana or give me the necessary permissions..

I've granted you additional permissions to proceed here (as I thought it is already configured -- both events are written to the same organization in Influx, just different buckets. In Grafana, there is a "default" bucket where I thought more access is possible. But maybe I restricted that when creating the token for the Grafana dashboard)

mpass99:
Thanks! I guess the access from Grafana to InfluxDB was ok. The existing CodeOcean Data Source uses InfluxQL where "codeocean" was specified as database (/ bucket). Which is why the other buckets were isolated..
Speaking of which, now that we're already creating a new dashboard, we're also using the newer language, Flux, right?

MrSerth:
Yes, using Flux is a good idea as we are creating a new dashboard anyway :)

mpass99:
How about using "Grafana as code" [[1]](https://grafana.com/blog/2020/02/26/how-to-configure-grafana-as-code/)?

| Pro                                                                                                   | Contra                                                                                                                                                                                                                                                                             |
|-------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------|
| Version Control (Possible Rollback, Reviewing, Teamwork)                                              | This is a very young area [[2]](https://grafana.com/go/grafanaconline/2021/dashboards-as-code/). Many tools are still in the beta phase [[3]](https://github.com/weaveworks/grafanalib) [[4]](https://github.com/grafana/grafonnet-lib) [[5]](https://github.com/grafana/grizzly). |
[...]

PR #113, opened 2022-04-09
Influxdb monitoring

### LINKED ISSUE #67: Extend statistics
https://github.com/openHPI/poseidon/pull/65

> > I assume that we won't send anything to Sentry in this case, right?
> 
> Yes, based on a normal termination, I would not have sent anything to Sentry; although it would be exciting to see how often this occurs. Should we rather log it at "warning" level?

> I see the interest but would rather not do it (at least not via Sentry). For one, Sentry is more dedicated to error tracing, and (therefore) the pricing model is based on each event. As a consequence, "counting" via Sentry might get expensive and as a project, we have an overall error limit per month (for all projects). Do you have another idea to log / count?

> We already have a "statistics" route. We could create a statistics controller and count things like "How often is which route requested" or "How often is the error code missing". With such a controller we could also provide an prometheus endpoint.
However, as this is a bit of work for currently not that much added value, I would assign it a low priority.

> I agree. Let's move this discussion into a dedicated issue and add it with a low priority. For these, we might also need to check for persistence/server restarts (depending on the concrete implementation).

mpass99:
Closes #67

What do you think about this first architecture?

We can add more information for the specific routes. However, I would suggest an interface change for easier data management: Adding `Environment() dto.EnvironmentID` to the Runner Interface. This way we can get the environment information just by having a runner (as we do in many routes).


MrSerth (approved):
Looks fine for me and is clean, I also think the proposed architectural change is something we should do. 👍

The only thing I was wondering about is passing the influxClient to the router, as this is opposed to the Sentry SDK or other logging mechanisms. However, in this context and for enriching the requests, it might be well-suited.

mpass99:
> this is opposed to the Sentry SDK or other logging mechanisms

In general, we try to [avoid global variables](https://medium.com/higher-order-functions/golang-its-a-global-problem-1a939461c16d).
For our logging system, we have chosen to make an exception.
Sentry has the global state in the client library itself.
For influxdb, We only need the influxdb client in one place - the middleware - unlike the logging system where the logging client could be needed anywhere.
That's why I decided against another global state in the form of the influxdb client.
But if you see it differently, that could easily be changed.

MrSerth:
> > this is opposed to the Sentry SDK or other logging mechanisms
> 
> In general, we try to [avoid global variables](https://medium.com/higher-order-functions/golang-its-a-global-problem-1a939461c16d). For our logging system, we have chosen to make an exception. Sentry has the global state in the client library itself. For influxdb, We only need the influxdb client in one place - the middleware - unlike the logging system where the logging client could be needed anywhere. That's why I decided against another global state in the form of the influxdb client. But if you see it differently, that could easily be changed.

Alright, that makes it very clear. I also agree that having as few global variables as possible is great because this makes using the object more explicit and reduces some unintended side effects. Hence, I fully support your decision 🤠

mpass99:
I have now added the environment id to the influxdb monitoring. Even if I still think that introducing `Environment() dto.EnvironmentID` was a good move, we still cannot access the environment data from the runner manager that well.
Now we would theoretically have all the data, but in practice we have isolated the runner manager so that it can only request the next runner manager in the chain of responsibility, but not the next environment manager.
This isolation could be removed, but I also understand the idea that the environment management (aka chain of responsibility) should only be used via the environment manager.
Should we rethink our architecture with the environment and runner manager or do we send the influxdb events without environment data (except the environment id) for the runner routes?

MrSerth:
Thanks for continuing this journey and adding `Environment()`. I started thinking of your comment as early as yesterday evening and immediately starting digging into the code (helping me to understand the problem you faced). Originally, I wondered why you (only) returned `dto.EnvironmentID` when calling `Environment()` but that got clearer when reading more code. I also tried myself to fetch more details with the `Get(id dto.EnvironmentID, fetch bool) (runner.ExecutionEnvironment, error)` method (as an external request would do) but unsurprisingly failed. Still, I would be interested to learn about your opinion on doing so: My intention was (as you probably tried as well), that this method would solve this issue but (as you pointed out in your previous comment):

> Now we would theoretically have all the data, but in practice we have isolated the runner manager so that it can only request runner functionality of the next runner manager in the chain of responsibility, but not the environments of the next runner manager.

The following question you're raising comes down to a central architectural decision you made early in the development of Poseidon: Having two distinct managers (one for the environment, and one for the runner) as opposed to one "combined" manager for both (which we had for a long time). 

> Should we rethink our architecture with the environment and runner manager or do we send the influxdb events without environment data (except the environment id) for the runner routes?

That's probably the right question. I assume that some form of connection (e.g., connect the runner manager to the respective environment [manager]) might be easier ideated than implemented. Hence, I also checked the requirements of the issue #67 we're doing here. Am I right that the "missing" data would be the environment type (Nomad vs AWS) and the number of used / idle runners in total (based on the checklist)? That might be sad but is not dramatic, we can live without these. Still, I would be curious about the number of used / idle runners (as this could make great visualisations in Grafana). What if we make some other architectural choices, e.g. some chain of responsibility for enriching the influx events? Otherwise, I might be tempted to continue without these data for now and maybe tackle them later on again.

mpass99:
It always makes me happy to have you on this journey :)

> The following question you're raising comes down to a central architectural decision you made early in the development of Poseidon

Yes, this is related to [the issue we closed recently](https://gitlab.hpi.de/codeocean/codemoon/poseidon/-/issues/62)

> fetch more details with the `Get(id dto.EnvironmentID, fetch bool) (runner.ExecutionEnvironment, error)` method

Following your train of thought has again revealed to me new approaches towards a solution.
Problem: Although we have the data, we cannot access it without the environment manager. Why don't we have the manager? Because we are in runner routes, i.e. in the runner controller. But our middleware is not in the runner controller... Why don't we give the middleware the possibility to look up its data itself?

We had to be careful with cyclic dependencies, but then the requirements of Issue #67 are fulfilled with this commit. What do you think of this way?

MrSerth:
Wow, that's what I imagined a few days ago (and couldn't achieve)! It's really fun to collaborate with you and see you implement so useful things that quickly 🤗

While you (of course) slightly changed the influxdb middleware (and increase the integration with the API), I think this is still separate enough to encapsulate the business logic (I hope you think so too). I also thought of the exact same usage of `Get` without fetching data (as you did). When reading the code, I (of course) did not think of any cyclic dependencies but it looks fine so far. However, it might be more tricky to add influxdb events for the environment routes, especially the Get route, I assume? (Still, this is not as important as the runner routes due to the different usage patterns).

> What do you think of this way?

I like it and the general approach you are going here; this should really make it simple to adopt the middleware in the near future or extend it. Also, I am glad about the custom-tailored explanation you provided for me 😊

mpass99:
> However, it might be more tricky to add influxdb events for the environment routes, especially the Get route, I assume?

Yes, you are right that we have to be careful about cyclic dependencies. But in principle I don't see any problem, because we mainly communicate via the request context.
Let's tackle this in a Follow-Up Issue of #110

PR #115, opened 2022-04-13
Add trace statements for WebSocket messages

MrSerth:
This MR adds the first messages for the most detailed log level available (the TRACE level). It is intended to debug application responsiveness in case of delays within the WebSocket connection. Therefore, each WebSocket message is logged with the corresponding timestamp. 

Due to the excessive logging of raw data, this log level is not recommended for production use but rather for local development.

### DIFF
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -117,6 +117,8 @@ func handleInput(messageType int, reader io.Reader, err error, buffer chan byte,
 		log.WithError(err).Warn("error while reading WebSocket message")
 		return true
 	}
+
+	log.WithField("message", string(message)).Trace("Received message from client")
 	for _, character := range message {
 		select {
 		case <-ctx.Done():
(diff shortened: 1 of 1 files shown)


internal/api/websocket.go
```diff
@@ -117,6 +117,8 @@ func handleInput(messageType int, reader io.Reader, err error, buffer chan byte,
 		log.WithError(err).Warn("error while reading WebSocket message")
 		return true
 	}
+
+	log.WithField("messageType", messageType).WithField("message", string(message)).Trace("Received message from client")
```
mpass99, line 121:
```suggestion
	log.WithField("message", string(message)).Trace("Received message from client")
```

We know that `messageType == websocket.TextMessage` otherwise we would return in line 112.

mpass99 (approved):

internal/api/websocket.go
MrSerth: Right, that's true. Thanks!

PR #116, opened 2022-04-18
Temporarily fixed golang version mismatch

mpass99:
in the ci build docker image step.

### DIFF
--- a/.github/workflows/docker-image.yml
+++ b/.github/workflows/docker-image.yml
@@ -21,7 +21,7 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v2
         with:
-          go-version: 1.16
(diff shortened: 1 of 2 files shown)


MrSerth (approved):
Looks good, thanks! Why is that only temporarily? (Because another mismatch might occur when not upgrading correctly?)

mpass99:
> Why is that only temporarily? (Because another mismatch might occur when not upgrading correctly?)

Yes, from Go 1.19 in four months, it can potentially occur again.
I have not yet researched possibilities to dynamically bind the Go version to the version defined in the Go module or other measures.

MrSerth:
Okay, good to know. If I do the next upgrade, I'll try to remember this MR...

PR #117, opened 2022-04-18
Fix typo in configuration.example.yaml

mpass99:
Unfortunately, this led to two additional production Poseidon restarts.

### DIFF
--- a/configuration.example.yaml
+++ b/configuration.example.yaml
@@ -75,6 +75,6 @@ influxdb:
   # The token to secure the influxdb endpoint.
   token: SecretBase64Token==
   # The organisation set in your influxdb configuration.
-  organisation: PoseidonOrg
+  organization: PoseidonOrg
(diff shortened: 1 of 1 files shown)


configuration.example.yaml
```diff
@@ -75,6 +75,6 @@ influxdb:
   # The token to secure the influxdb endpoint.
   token: SecretBase64Token==
   # The organisation set in your influxdb configuration.
```
MrSerth, line 77: Let's fix it in the string as well. Otherwise, we might get confused with the American and British versions ...

MrSerth (approved):
Don't worry, these mistakes happen. I am glad you found the root cause and fixed that already ;)

configuration.example.yaml
mpass99:
```suggestion
  # The organization set in your influxdb configuration.
```

MrSerth (approved):

PR #119, opened 2022-04-18
Allow the CR character as part of a newline.

mpass99:
Since CodeOcean may send the `\r\n` newline.


MrSerth (approved):
Yes, that is (unfortunately) not always `\n`. I also now that receiving data through Poseidon (in the websocket connection) might include `\r\n` (e.g., when it is received from the Python environment). 

👏  Great to see a regression test - thanks!

Issue #124, opened 2022-04-20
AWS: Tests cannot get executed


MrSerth:
I tried the AWS Hello World again and noticed that the `Score` functionality is currently broken.

Instead of executing the tests, it simply returns:

```
StdErr: sh: make: command not found
```

Also, in CodeOcean, we should switch to JUnit 4 (from JUnit 5) for the AWS environment (as this is the version the tests are written in).

mpass99:
This is because our Makefile implementation does not recognise `make test CLASS_NAME="%{class_name}" FILENAME="%{filename}"` as valid make command matching the pattern `^make(?:\s+(?<startRule>\w*))?$`.

Should we implement the make feature of make variables? We should also keep in mind that there are other possibilities, such as putting a static make binary (+ dependencies if necessary) into the aws deployment.

MrSerth:
Hm, that's another edge case for Make, true. I would hope, it is the last one...

Regarding the tests: Do you know whether they will work once we fixed that make problem? I am asking as we are also including JUnit 4 with two jars which might or might not be available in AWS. Probably using JUnit 5 is no problem, but any JUnit is required.

mpass99:
Regarding the tests: We can include all dependencies via our maven [pom.xml](https://github.com/openHPI/poseidon/blob/main/deploy/aws/java11Exec/pom.xml) file.

For example, if we remove the test-scope of our junit dependency...

https://github.com/openHPI/poseidon/blob/bb2d989bc51680a29e8c7ebb1ea014ae647b32ac/deploy/aws/java11Exec/pom.xml#L35-L40

... we can access `junit.junit-4.13.1.jar` in the folder `/var/task/lib`.
How do we deal with the problem that the path to the library is set statically in the Makefile, but differs between Nomad and AWS?

MrSerth:
Okay, that's fine. We use that dependency and Hamcrest (in addition):

```
<!-- https://mvnrepository.com/artifact/org.hamcrest/hamcrest-core -->
<dependency>
    <groupId>org.hamcrest</groupId>
    <artifactId>hamcrest-core</artifactId>
    <version>1.3</version>
</dependency>
```

My proposal (for a quick solution, but not tested yet):

```
.map(s->s.replaceAll("/usr/java/lib/hamcrest-core-1.3.jar", "/<new path>.jar"))
.map(s->s.replaceAll("/usr/java/lib/junit-4.11.jar", "/<new path>.jar"))
```

[...]

PR #131, opened 2022-04-23
Java11Exec: Add support for make assignments.

mpass99:
Solves first part of #124

### DIFF
--- a/deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java
+++ b/deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java
@@ -7,6 +7,7 @@ import java.util.HashMap;
 import java.util.Map;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
+import java.util.stream.Collectors;
 
 // NoMakefileFoundException is thrown if no makefile could be found.
 class NoMakefileFoundException extends Exception {}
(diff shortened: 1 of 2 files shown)


deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java
```diff
@@ -21,7 +22,7 @@ class InvalidMakefileException extends Exception {}
 class SimpleMakefile {
 
     // This pattern validates if a command is a make command.
-    private static final Pattern isMakeCommand = Pattern.compile("^make(?:\\s+(?<startRule>\\w*))?$");
+    private static final Pattern isMakeCommand = Pattern.compile("^make(?:\\s+(?<startRule>\\w*))?(?<assignments>(?:.*?=\\\".*?\\\")+)?$");
```
MrSerth, line 25: I am not sure whether the `"` are required (or optional), but otherwise this MR looks fine. Thanks again ;)

MrSerth (dismissed):

deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java
mpass99:
```suggestion
    private static final Pattern isMakeCommand = Pattern.compile("^make(?:\\s+(?<startRule>\\w*))?(?<assignments>(?:.*?=.*?)+)?$");
```

MrSerth: If we remove them, do we need to adjust the slicing here as well?

https://github.com/openHPI/poseidon/blob/a733daffc8644fe83e389c2c25bfcd11e63dab9c/deploy/aws/java11Exec/src/main/java/poseidon/SimpleMakefile.java#L94

mpass99: Good catch, thanks

MrSerth (approved):
Smart solution, I like that :+1:

PR #132, opened 2022-04-24
Java11Exec: Quick fix for test dependencies.

### LINKED ISSUE #124: AWS: Tests cannot get executed

mpass99:
Closes #124


MrSerth (approved):
Looks good, I would suggest to give it a try ;)

Issue #89, opened 2022-01-20
Refactoring: Use generics for the object storages


mpass99:
With [nomad_environment_storage.go](https://github.com/openHPI/poseidon/blob/main/internal/runner/nomad_environment_storage.go), [storage.go](https://github.com/openHPI/poseidon/blob/main/internal/runner/storage.go) and [local_storage.go](https://github.com/openHPI/poseidon/blob/main/pkg/execution/local_storage.go) we have three kinds of saving objects in a dictionary. All are implemented separately and have their own test file. 

We may use golang generics introduced in Go 1.18 (02/2022) to reduce this to only one local storage.

PR #145, opened 2022-05-24
Add Grafana Dashboard

### LINKED ISSUE #110: Create Grafana Dashboards

mpass99:
including information about the number of requests, runner and executions and their durations.

What do you think? Is all the important information visualized? Could the structure of the dashboard be improved?

One visualization I am missing is how much time it takes Nomad to start the runners. This information can be added in the runner manager but it is not bound to a request of CodeOcean. This problem applies also for the statistics `idle_runner` and `prewarming_pool_size` - we send the same data with all incoming requests but not at the moment the change happens. That's the Prometheus way but not the Influx'... Maybe we should reconsider our decision to restrict the access to the influx client and to not provide it as singleton?!

I have no idea how to normalise the histogram.

### DIFF
--- a/internal/environment/aws_environment.go
+++ b/internal/environment/aws_environment.go
@@ -87,7 +87,8 @@ func (a *AWSEnvironment) CPULimit() uint {
 func (a *AWSEnvironment) SetCPULimit(_ uint) {}
 
 func (a *AWSEnvironment) MemoryLimit() uint {
-	return 0
+	const memorySizeOfDeployedLambdaFunction = 2048 // configured /deploy/aws/template.yaml
+	return memorySizeOfDeployedLambdaFunction
 }
 
 func (a *AWSEnvironment) SetMemoryLimit(_ uint) {
@@ -95,7 +96,7 @@ func (a *AWSEnvironment) SetMemoryLimit(_ uint) {
 }
 
 func (a *AWSEnvironment) NetworkAccess() (enabled bool, mappedPorts []uint16) {
-	return false, nil
+	return true, nil
 }
 
 func (a *AWSEnvironment) SetNetworkAccess(_ bool, _ []uint16) {
--- a/internal/environment/aws_environment.go
+++ b/internal/environment/aws_environment.go
@@ -87,7 +87,7 @@ func (a *AWSEnvironment) CPULimit() uint {
 func (a *AWSEnvironment) SetCPULimit(_ uint) {}
 
 func (a *AWSEnvironment) MemoryLimit() uint {
-	panic("not supported")
+	return 0
 }
 
 func (a *AWSEnvironment) SetMemoryLimit(_ uint) {
@@ -95,7 +95,7 @@ func (a *AWSEnvironment) SetMemoryLimit(_ uint) {
 }
 
 func (a *AWSEnvironment) NetworkAccess() (enabled bool, mappedPorts []uint16) {
-	panic("not supported")
+	return false, nil
 }
 
 func (a *AWSEnvironment) SetNetworkAccess(_ bool, _ []uint16) {
--- a/internal/environment/nomad_environment.go
+++ b/internal/environment/nomad_environment.go
@@ -11,6 +11,7 @@ import (
 	"github.com/openHPI/poseidon/internal/nomad"
 	"github.com/openHPI/poseidon/internal/runner"
 	"github.com/openHPI/poseidon/pkg/dto"
+	"github.com/openHPI/poseidon/pkg/monitoring"
 	"github.com/openHPI/poseidon/pkg/storage"
 	"strconv"
 	"sync"
@@ -35,7 +36,8 @@ func NewNomadEnvironment(apiClient nomad.ExecutorAPI, jobHCL string) (*NomadEnvi
 		return nil, fmt.Errorf("error parsing Nomad job: %w", err)
 	}
 
-	return &NomadEnvironment{apiClient, jobHCL, job, storage.NewLocalStorage[runner.Runner]()}, nil
+	return &NomadEnvironment{apiClient, jobHCL, job,
+		storage.NewMonitoredLocalStorage[runner.Runner](monitoring.MeasurementIdleRunnerNomad, nil)}, nil
 }
 
 func NewNomadEnvironmentFromRequest(
@@ -80,6 +82,7 @@ func (n *NomadEnvironment) PrewarmingPoolSize() uint {
 }
 
 func (n *NomadEnvironment) SetPrewarmingPoolSize(count uint) {
+	monitoring.ChangedPrewarmingPoolSize(n.ID(), count)
 	taskGroup := nomad.FindAndValidateConfigTaskGroup(n.job)
 
 	if taskGroup.Meta == nil {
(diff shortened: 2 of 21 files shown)


MrSerth:
I really like the visualizations! They are so nice to look at (and contain much information), will definitely recommend them! 
⭐ ⭐ ⭐ ⭐ ⭐ 

Some suggestions:
- Length of the smallest bar for the bar charts should be increased
- Naming of environments: could we add a mapping, even if done manually once?

Regarding your questions:

> One visualization I am missing is how much time it takes Nomad to start the runners. This information can be added in the runner manager but it is not bound to a request of CodeOcean. This problem applies also for the statistics idle_runner and prewarming_pool_size - we send the same data with all incoming requests but not at the moment the change happens. 

Wouldn't it be possible to send these events anyway as soon as they occur, even outside of the usual request-response cycle?

> Maybe we should reconsider our decision to restrict the access to the influx client and to not provide it as singleton?!

Fine for me ;)

> I have no idea how to normalise the histogram.

I think it is fine with the current visualization. I tried a few things in Excel, but unfortunately was unable to proceed. First, Excel does not allow multiple data sources in one histogram (as required for the multiple execution environments). Also, I was unable to find a good way of normalizing them individually. Therefore, and as the height also includes some information, I would suggest keeping it with the current visualization for now. We might only consider adding an overall gauge of the total usage of Poseidon, as discussed earlier today ;)

mpass99:
> Wouldn't it be possible to send these events anyway as soon as they occur, even outside of the usual request-response cycle?

Yes, for example we could notify influx for every created or deleted object in our storage. This would include `idle_runners`, `used_runners`, `aws_executions`, `nomad_executions`, `environments`. To do so I would start #89 now.

MrSerth:
> > Naming of environments: could we add a mapping, even if done manually once?
> 
> I could not find a Grafana feature for this (as we do not want to map values but labels), but we could map this via Flux and an additional data source (as found in some blog post). Should we therefore install [the CSV plugin](https://github.com/marcusolsson/grafana-csv-datasource) or use a database for the manual mapping between environment ID and environment name?

Yes, that might work. Alternatively, we could also wait for #89 and those additional events. Then, we could send those "static" information once, e.g., when synchronizing an execution environment and just send the name together with the desired prewarming pool size, RAM, ... to Influx. As they can only be changed with an update / synchronisation, this could also help us to build a corresponding data store.

MrSerth:
Really great dashboards, I just used them again. ~~One~~ Two final questions: 
- Could we increase the aggregation for the line charts if a window larger >= 6hours is seen? In that case, we might want to switch to 5 min instead of 1 min.
- Could we use the full page width for the first histogram "Execution Duration" in order to make that easier to use?

mpass99:
> Then, we could send those "static" information once, e.g., when synchronizing an execution environment and just send the name together with the desired prewarming pool size, RAM, ... to Influx

Update Environment information in extra measurement with filter for one year.

mpass99:
> Then, we could send those "static" information once, e.g., when synchronizing an execution environment and just send the name together with the desired prewarming pool size, RAM, ... to Influx. As they can only be changed with an update / synchronisation, this could also help us to build a corresponding data store.

With the new commit the data should be saved in another measurement. In the next pull request we could adjust the Grafana dashboard to create the mapping in order to display the environment image as name..

What else is an open ToDo here:
- collect data to measure how long Nomad takes to start an allocation?
-

internal/environment/aws_environment.go
```diff
@@ -87,15 +87,15 @@ func (a *AWSEnvironment) CPULimit() uint {
...
+	return 0
 }
 
 func (a *AWSEnvironment) SetMemoryLimit(_ uint) {
 	panic("not supported")
 }
 
 func (a *AWSEnvironment) NetworkAccess() (enabled bool, mappedPorts []uint16) {
-	panic("not supported")
+	return false, nil
```
MrSerth, line 98: I would rather switch this flag to `true`.

In my latest experiment, I was able to access the internet using AWS:

```
import java.net.*;
import java.io.*;
import java.util.*;

public class HalloWelt {
    public static void main (String [] args) throws MalformedURLException, IOException {
        System.out.println("Hallo Internet!");
        String out = new Scanner(new URL("http://www.google.com").openStream(), "UTF-8").useDelimiter("\\A").next();
        System.out.println(out);
    }
}
```

internal/environment/aws_environment.go
```diff
@@ -87,15 +87,15 @@ func (a *AWSEnvironment) CPULimit() uint {
 func (a *AWSEnvironment) SetCPULimit(_ uint) {}
 
 func (a *AWSEnvironment) MemoryLimit() uint {
-	panic("not supported")
+	return 0
```
MrSerth, line 90: Shall we stay with 0 here or switch to 1536 MB according to our deployment (and [analysis](https://lambda-power-tuning.show/#AAIABAAGAAgADA==;3md0RfaQ5EQw0pZEKfRzRBsQNkQ=;5hffN16p0DcPjc43rrHeNwmB+Tc=;AAIABAAGAAgADA==;k/qFRRucAUUsBaZEcV2ERNfTNUQ=;UhMXOJokEjh5eAw4UD4VOODkGTg=;ARM64;x86_64))?

internal/environment/nomad_environment.go
```diff
@@ -80,6 +82,7 @@ func (n *NomadEnvironment) PrewarmingPoolSize() uint {
 }
 
 func (n *NomadEnvironment) SetPrewarmingPoolSize(count uint) {
+	monitoring.ChangedPrewarmingPoolSize(n.ID(), count)
```
MrSerth, line 85: Is it okay to have the monitoring line here or shall we move it below the next line(s) to ensure we have a valid job config and don't log any wrong data?

MrSerth (dismissed):
Looks good to me :shipit: 
I reviewed all Go-related changes but left out the Grafana dashboard JSON -- that is much easier to review once it is deployed ;)

internal/environment/aws_environment.go
mpass99: Yes, nice :) Currently we use 2048MB that is not cost optimized (as described by the analysis) but with manageable costs of 0,52$ in the last 6 months.
https://github.com/openHPI/poseidon/blob/344dfdfe58fe06b7739c7c65f34b52853f81558b/deploy/aws/template.yaml#L66

```suggestion
	return 2048
```

internal/environment/nomad_environment.go
mpass99: Since the check in the next lines could just lead to initialization of the data structure and not to an invalidation of the job or the data, I think its makes not difference..

MrSerth: Alright, fine for me!

internal/environment/aws_environment.go
MrSerth: Okay, these costs are really acceptable 😁

MrSerth (approved):

PR #147, opened 2022-05-24
Set and unset aws environment variables

### LINKED ISSUE #130: Java11Exec: Visible credentials
The AWS lambda function contain the credentials of the specific lambda execution role as environment variables. With `env` you can read them and use them for requests with your AWS cli. The iam role for the lambda function has not many permissions but I was able to create Logs (and Log groups) via those credentials.

As a measure we can on the one hand hide all environment variables during execution and on the other hand remove the permission to write logs, as we don't use the logs anyway.

mpass99:
in order to provide the CodeOcean context for some scripts and hide AWS credentials from the users.

Closes #130


internal/runner/aws_runner.go
```diff
@@ -208,3 +209,12 @@ func (w *AWSFunctionWorkload) handleRunnerTimeout(ctx context.Context,
...
 	}
 }
+
+// prepareEnvironmentVariables sets the CODEOCEAN variable and unsets some AWS variables.
+func (w *AWSFunctionWorkload) prepareEnvironmentVariables(request *dto.ExecutionRequest) {
+	if request.Environment == nil {
+		request.Environment = make(map[string]string)
+	}
+	request.Environment["CODEOCEAN"] = "true"
+	request.Command = "unset \"${!AWS@}\" && " + request.Command
```
MrSerth, line 219: That's clever and works perfectly fine in Bash. I just tested that locally with a ZSH, in which it does not work. As we only specify "sh" as a command-line interpreter, are we sure this won't break unnoticed?

mpass99: I have now introduced an e2e test for this. Thanks for your good tip!

MrSerth (approved):
Great, thanks for adding an e2e test!

PR #149, opened 2022-05-25
Memory oversubscription documentation

### LINKED ISSUE #139: Document Memory Oversubscription
So far, our documentation lacks details on the memory oversubscription. If that is not enabled, allocations are very likely to fail. We should add it [here](https://github.com/openHPI/poseidon/blob/main/docs/configuration.md).

mpass99:
Closes #139


MrSerth (approved):
Great, thanks!

PR #150, opened 2022-05-28
Specify AWS Functions as list

### LINKED ISSUE #95: Specify AWS Functions as list
Multiple functions are separated by space in a string? What about using a list type?
```yml
functions:
- java11Exec
- go118Exec
```

_Originally posted by @MrSerth in https://github.com/openHPI/poseidon/pull/87#discussion_r803125714_

Yes that would be cleaner. I had decided against this, because for the e2e tests we also want to set this configuration as an environment variable (see .github/workflows/ci.yml).
Another solution would be to insert the case for `reflect.Slice` [here](https://github.com/openHPI/poseidon/blob/1d974d9576ef0a915682c29b1811a93e63fe4c86/internal/config/config.go#L184-L213). Would you prefer this solution?

_Originally posted by @mpass99 in https://github.com/openHPI/poseidon/pull/87#discussion_r803661319_

mpass99:
to conform with the yaml standard of list definition.

Closes #95

### DIFF
--- a/configuration.example.yaml
+++ b/configuration.example.yaml
@@ -48,8 +48,11 @@ aws:
   enabled: false
   # The endpoint of the WebSocket API
   endpoint: wss://abcdef1234.execute-api.eu-central-1.amazonaws.com/production
-  # Currently, only static AWS environments are supported. You can list them here separated by spaces.
-  functions: "java11Exec go118Exec"
+  # Currently, only static AWS environments are supported.
+  # For setting this via environment variables you have to use a string separated by spaces, like: POSEIDON_AWS_FUNCTIONS="java11Exec go118Exec".
+  functions:
+    - java11Exec
+    - go118Exec
 
 # Configuration of the logger
 logger:
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -47,7 +47,7 @@ var (
 		AWS: AWS{
 			Enabled:   false,
 			Endpoint:  "",
-			Functions: "",
+			Functions: []string{},
 		},
 		Logger: logger{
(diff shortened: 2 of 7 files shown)


configuration.example.yaml
```diff
@@ -48,8 +48,11 @@ aws:
   enabled: false
   # The endpoint of the WebSocket API
   endpoint: wss://abcdef1234.execute-api.eu-central-1.amazonaws.com/production
-  # Currently, only static AWS environments are supported. You can list them here separated by spaces.
-  functions: "java11Exec go118Exec"
+  # Currently, only static AWS environments are supported.
+  # For setting this via environment variables you have to use a string separated by spaces, like: POSEIDON_AWS_FUNCTIONS="java11Exec go118Exec".
```
MrSerth, line 52: I assume this comment is now outdated, isn't it? 🤓

MrSerth (approved):
Looks good, there is just a minor remark regarding a comment 🦉

configuration.example.yaml
mpass99: Hm, I guess not. Since [`The value of an environment variable is a string of characters`](https://pubs.opengroup.org/onlinepubs/9699919799/basedefs/V1_chap08.html) the option to set this value as environment variable (not via the config file! Like it is used in our CI) must still rely on a string separated by any kind of delimiter..

This PR just fixes that our yaml config file is parsing a list from a string although lists can be specified in yaml.

MrSerth: Ah, I see. 🙈 Thanks for reminding me 😄

PR #158, opened 2022-06-01
Fix CodeQL log injection warning

### LINKED ISSUE #51: Fix security alerts

mpass99:
by removing newlines from logged user input.

Closes #51


MrSerth (dismissed):
I am somewhat surprised that removing the newline characters is enough for this rule, but if that is the case, I am happy 😃

PR #159, opened 2022-06-02
File Transfer back from runner

### LINKED ISSUE #136: Enable file transfer from allocations to CodeOcean
So far, we have a one-way file copy mechanism, which allows CodeOcean to send files to an allocation. These files usually contain the user's source code or some other resources required for the execution.

In some circumstances, we are also interested in receiving files generated during the execution within the allocation and would like to receive them. For example, a user in the R course might use R to generate a graph using code and some input data, which is consequently exported as a PDF file. This PDF file is generated within the allocation and, so far, we don't have any mechanism to provide the user with a download link to this PDF. Therefore, I suggest investigating how we may receive these artifacts.

As a user,  
I want to receive the files generated with my source code   
in order to download and review them on my local machine.

<hr>

In order to implement this feature, we could design an additional API for Poseidon that CodeOcean uses. Ideally, we have two routes: One for receiving a *list* of files (and all sub-directories) in `/workspace` (if not specified differently) with some basic information (such as the file name incl. file extension, file size, maybe modification date) and another one for retrieving the actual *content* of files (with an option to receive either all files or only those selected). When offering the files for download (in CodeOcean), we should also remove any file that was sent to Poseidon previously (as that might contain hidden test script files) or derived from a file we sent (such as `.class` files for matching `.java` files). Ideally, the list of all files would have an additional flag whether the file was created by the user or us, but if that is too hard, we just filter them by name in CodeOcean).

As this is a quite complex feature, we might start with designing the API and some MVP to validate the feasibility of this request.

mpass99:
Closes #136


api/swagger.yaml
```diff
@@ -312,6 +367,40 @@ paths:
...
+        - name: runnerId
+          description: Runner on which the command should be executed
+          in: path
+          schema:
+            type: string
+            example: 123e4567-e89b-12d3-a456-426614174000
+          required: true
+        - name: filename
+          in: path
+          description: Specify the file that should be returned by it's filename including it's path and it's extension.
```
MrSerth, line 386:
```suggestion
          description: Specify the file that should be returned by its filename including its path and extension.
```

api/swagger.yaml
```diff
@@ -312,6 +367,40 @@ paths:
         "500":
           $ref: "#/components/responses/InternalServerError"
 
+  /runners/{runnerId}/files/{filename}:
```
MrSerth, line 370: Would it be possible to acquire a file located at `../flag.txt`? Or is that not a valid use case for our system?

MrSerth (commented):
I like the API (and examples), seems to be a good start 😊

api/swagger.yaml
MrSerth: I find both approaches valid, but would go with the one I prefer: Using the `/runners/{runnerId}/files/raw` URL and a query parameter. This does not mingle the response scheme and makes it easier to extend the use case later (think of a zip compressed file download service, located under `/runners/{runnerId}/files/zipped`, for example).

api/swagger.yaml
```diff
@@ -312,6 +367,40 @@ paths:
...
+          schema:
+            type: string
+            example: 123e4567-e89b-12d3-a456-426614174000
+          required: true
+        - name: filename
+          in: path
+          description: Specify the file that should be returned by it's filename including it's path and it's extension.
+          schema:
+            type: string
+            default: "./flag.txt"
```
MrSerth, line 389: That's rather an example than a default, isn't it?

MrSerth:
Awesome, this works great! 💪 I just tested it with a PDF I generated within a runner and even saw a preview in Postman (the Content-Type was set correctly):

Bildschirmfoto 2022-08-16 um 18 37 43

Just one remark: The download function currently logs one of our already know warning messages, is there something to adjust?
```log
2022/08/16 18:41:13 http: superfluous response.WriteHeader call from github.com/openHPI/poseidon/pkg/logging.(*loggingResponseWriter).WriteHeader (logging.go:59)
time="2022-08-16T18:41:13.743877Z" level=debug code=200 duration=160.839ms method=GET path=/api/v1/runners/28-142643b0-1d7f-11ed-89f8-acde48001122/files/raw user_agent=PostmanRuntime/7.29.0
```

Besides that, I also tried listing files, but that did not work as expected:

```bash
curl -v http://localhost:7200/api/v1/runners/28-1427640c-1d7f-11ed-89f8-acde48001122/files
*   Trying 127.0.0.1:7200...
* Connected to localhost (127.0.0.1) port 7200 (#0)
> GET /api/v1/runners/28-1427640c-1d7f-11ed-89f8-acde48001122/files HTTP/1.1
> Host: localhost:7200
> User-Agent: curl/7.79.1
> Accept: */*
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 405 Method Not Allowed
< Date: Tue, 16 Aug 2022 16:46:30 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
*
```

Is this feature already available as described in the swagger.yaml?

mpass99:
Wow, thanks for catching so many issues / improvements!!

1. Yes, `path` and all query parameters should be URI-encoded. Should we add a comment for this. Postman makes this automatically and some websites state that URI-encoding is the default for query parameters.
The json response should be fixed. Also, the cause of the internal server error should be fixed: The url encoded ampersand broke it in the way that it was not enquoted by Poseidon for the shell execution.

MrSerth:
> Wow, thanks for catching so many issues / improvements!!

You're welcome, thanks for taking the time to work on the improvements!

> 1. Yes, `path` and all query parameters should be URI-encoded. Should we add a comment for this. Postman makes this automatically and some websites state that URI-encoding is the default for query parameters.

Yes, that is now working as expected. Using Postman, many of the special characters get encoded correctly, just the percentage sign `%` needs to be encoded manually (if that is literally included). I now checked the file listing and file retrieval with `'`, `"`, `&`, `%` as the most interesting cases. Maybe just add a comment (or is that clear otherwise because it is the default)? Would be fine for me as well otherwise.

>    The json response should be fixed. Also, the cause of the internal server error should be fixed: The url encoded ampersand broke it in the way that it was not enquoted by Poseidon for the shell execution.

Confirmed, is now working as expected.

> 2. I see the inaccuracy here. But, as "we don't need any differences here", what should we change? Do you want the error message not to be `file not found` for the file listing?

What if we make it explicit?
> File not found or insufficient permissions

Another question: As of now, we have an overlap of the "Runner not found" and "File not found" JSON response. Both render a 404 status code and "just" some different "message", that is not well machine readable. We could either add an additional errorCode enum or change the HTTP status code. For the HTTP status code, I think we could use different error codes, here is a selection of possible ones:

- 410 Gone (/ 417 Expectation Failed / 424 Failed Dependency): Return if the runner is not available any longer (instead of 404 for the runner)
- 424 Failed Dependency (/ 421 Misdirected Request): Return if the runner is available but the file is not (instead of 404 for the file).

All are not perfect, I assume the 410 Gone would be the best for a replacement of the runner 404.

mpass99:
Yes, that's a good suggestion! Thanks

mpass99:
3. I find the link target interesting :)
#
1. As CodeOcean already will filter the files, I would list all files here.
2. Yeah, I like to go with the `ls` wording.
3. You got me with "for free" ;)

MrSerth:
Awesome, this is really nice -- I cannot wait to get it live. We just need to ensure not to break CodeOcean, the change of the HTTP status codes need to be performed there as well (I haven't done so yet). For the new status codes and API responses, I added three minor comments to the commit https://github.com/openHPI/poseidon/commit/378848b8445d368874750554cab54183e0f2af47; please have a look there 🙂

MrSerth (commented):
Awesome work! I just tested it again and everything works as expected. I am just a little sad by PDF file is not rendered any longer correctly (based on the MIME type change that occurred earlier). Do we have a *simple* possibility to use the MIME type provided by the file and just fallback to `application/octet-stream` if it couldn't be determined or is not registered? Or do we don't want that behaviour intentionally?

Bildschirmfoto 2022-08-31 um 14 10 32

mpass99:
Yeah, this would be a nice to have.
In a small online check I found two commands that would be suitable: `file` and `xdg-mime` (file content vs. file extension analysis). Neither is installed in our Docker images. Is this feature worth adding one of them?
In a small PoC I implemented a solution, but I was torn between two ways and would be interested in your preference: In `NomadJob` `GetFileContent`...
- Execute two commands after each other (simple Code)
- Execute two commands in one request and split the outputs via a custom writer (faster).

pkg/nullio/content_type.go
```diff
@@ -0,0 +1,43 @@
...
+		return count, fmt.Errorf("could not write to target: %w", err)
+	}
+
+	for i, char := range p {
+		if char != '\n' {
+			continue
+		}
+
+		w.mimeType = append(w.mimeType, p[:i]...)
+		w.Target.Header().Set("Content-Type", string(w.mimeType))
```
MrSerth, line 28: We should also consider to add a [`Content-Disposition` header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition) with the `attachment` keyword (this is a must), and optionally provide the filename in the header as well.

MrSerth (commented):
Wow, thanks for starting to work on that! I didn't expect that note to have so many consequences as it worked before (compare [this comment (working)](https://github.com/openHPI/poseidon/pull/159#issuecomment-1216892940) against [this comment (not working any longer)](https://github.com/openHPI/poseidon/pull/159#pullrequestreview-1091784039)). Therefore, I (wrongly?) assumed that maybe explicitly setting the header with `writer.Header().Set("Content-Type", "application/octet-stream")` would cause this behaviour and that otherwise the framework would just _magically_ provide the correct content type (which would be similar to Ruby on Rails). Are you sure we cannot use such as default mechanism as before? Or do you have any idea why it used to work and now does no longer?

That being said, I also don't have any issue providing a small utility, such as the `file` command, in our containers if we continue with this approach. I also get the question you raised and actually like the current approach with concatenating both information (the mime-type as received through file and the file content). Sure, this is more more complex, but I would go for the (slightly) better performance here [and as I can see, you already did that, right?]). 

👏💪🚀

pkg/nullio/content_type.go
mpass99: Yeah, this would be an easy improvement :+1: 
More complex would be to set the `Content-Length` header for large files. For small files this is done automatically by Go:
``` Go
        // Additionally, if the total size of all written
	// data is under a few KB and there are no Flush calls, the
	// Content-Length header is added automatically.
```
For larger files, we could think about adding the Content-Length header the way [e1d16e5](https://github.com/openHPI/poseidon/pull/159/commits/e1d16e523b5c7bafc8485b843836388110eca253) has added (unnecessarily) the Content-Type?!

MrSerth: Sure - if you want, just do so. We already have the infrastructure anyway with your implementation and just need to swap the command and header 😁

MrSerth:
Once #215 is merged, we should also consider the user switch for retrieving files.

MrSerth (commented):
Both HTTP endpoints work fine, just as expected. 👍

~~However, I wondered that they are not protected through our Authorisation token (but just unauthenticated). I would vote to enable authorisation on these endpoints as well 😉~~ 
Edit: I disabled token auth in my logal config 🤦

MrSerth:
I worked on the integration of the file download feature in CodeOcean recently. While working on that, I noticed a few things:

**Blockers:**
- [ ] Token Authentication is not required for the endpoints
- [ ] The JSON response might be invalid, if requesting `/` with recursive `true` as a non-privileged user (the exit code is != 0 resulting in the error JSON being appended to the previous output JSON)

**Should have:**
- [ ] Add a `privilegedExecution` flag (or named differently) to the two routes for listing the file system and retrieving files. This should allow to specify the user which is needed to retrieve files.

**Nice to have:**
- [ ] Consider piping the `cat` output through a compression program in the container and decompress it in Poseidon to allow for faster data transfer? Is that possible with streaming? The goal would be to reduce the bytes sent through Nomad, as this lead to a very high load for Nomad with large files (I tested it 1G):

  ```ruby
  system("dd if=/dev/zero of=test.txt bs=1 count=0 seek=1G") # Can be compressed very well!
  system("dd if=/dev/random of=test.txt bs=1 count=0 seek=1G") # Cannot be compressed probably
  ```

mpass99:
> The JSON response might be invalid, if requesting `/` with recursive `true` as a non-privileged user (the exit code is != 0 resulting in the error JSON being appended to the previous output JSON)

The exit code is `1` because `ls` throws many errors like `ls: cannot open directory '/etc/container_environment': Permission denied` on stdErr.

I see two ways how to handle this:
1. Ignore the exit code.
2. Adjust the Swagger `FileHeader` to also allow just `name` and `error` as an valid return object. Extend the `Ls2JsonWriter` to handle also the `ls` stdErr output.

What do you think? Is the error message per file an useful information for CodeOcean?

MrSerth:
> What do you think? Is the error message per file an useful information for CodeOcean?

In my current implementation, I am not handling errors on a file level. Therefore, I would use something like your second proposal: We are listing all files currently accessible (in terms of "showing with ls") for the given user. Iff the path I transmitted to Poseidon is not accessible at all (such as `/root` for non-privileged users), I would expect an error code (!= 200) and that error JSON.

Adding the error information to the JSON (suggestion 3) would also be fine for me. This is something I would consider in addition to the previous proposal and think it is **optional**: If the result is filtered (such as due to the permission denied error), each failed entry could get an error. If possible, I would also keep the "entry type". How exactly would the response look like for `/root` or `/etc/container_environment` -- like this?

```json
    {
      "name": "/root",
      "entryType": "d",
      "size": 4096,
      "modificationTime": 1654201799,
      "error": "Permission denied"
    }
```

mpass99:
I have implemented the second option.
I would wait with the third one until a use case appears. A simple implementation would just use the information of the error message leading to:
``` json
    {
      "name": "/root",
      "entryType": "d",
      "error": "Permission denied"
    }
```

mpass99:
> Add a `privilegedExecution` flag (or named differently) to the two routes for listing the file system and retrieving files. This should allow to specify the user which is needed to retrieve files.

This should already work as described in the Swagger API.

MrSerth:
> I have implemented the second option. I would wait with the third one until a use case appears. A simple implementation would just use the information of the error message leading to:
> 
> ```json
>     {
>       "name": "/root",
>       "entryType": "d",
>       "error": "Permission denied"
>     }
> ```

Alright, I would agree with that structure.

Additionally, I am also fine to wait with the implementation until a use case arises.

MrSerth:
> > Add a `privilegedExecution` flag (or named differently) to the two routes for listing the file system and retrieving files. This should allow to specify the user which is needed to retrieve files.
> 
> This should already work as described in the Swagger API.

Awesome! I touched my implementation and can now confirm that it works as expected -- sorry again... 🙈 🙉 🙊

PR #163, opened 2022-06-09
#42 Allow deletion of ./*

### LINKED ISSUE #42: Unable to remove all files within a directory
In the current version, Poseidon sanitizes the path before executing the delete request in Nomad. This has been implemented to increase security and prevent shell escapes and is working as intended:

https://github.com/openHPI/poseidon/blob/cf3db036facf0b97f8bfd4598b59489550613a6f/internal/runner/runner.go#L270-L284

However, the sanitization prevents the usage through CodeOcean:

```ruby
    # First, clean the workspace, and second, copy all files to their location.
    # This ensures that no artifacts from a previous submission remain in the workspace.
    body = {copy: copy, delete: ['./']}
```

The specified filename gets changed to '.' which is not a valid path to remove:

```
rm: "." and ".." may not be removed
```

When changing the directory path to delete everything **within** the current path (as in `{copy: copy, delete: ['./*']}`), we issue the following command within Nomad after sanitizing it in Poseidon:

```
rm --recursive --force '*'
```

This is not working as intended, as the quotes change the semantic to delete a file named `*` and not everything within the current folder (as initially expected). Therefore, it is currently not possible to remove all files within a given directory without deleting the whole directory.

My workaround, for now, is to delete the outer folder `/workspace` including all subfiles. It works temporarily but is not the optimal solution.

mpass99:
Closes #42

### DIFF
--- a/internal/runner/nomad_runner.go
+++ b/internal/runner/nomad_runner.go
@@ -253,11 +253,15 @@ func fileDeletionCommand(pathsToDelete []dto.FilePath) string {
 	}
 	command := "rm --recursive --force "
 	for _, filePath := range pathsToDelete {
-		// To avoid command injection, filenames need to be quoted.
-		// See https://unix.stackexchange.com/questions/347332/what-characters-need-to-be-escaped-in-files-without-quotes
-		// for details.
-		singleQuoteEscapedFileName := strings.ReplaceAll(filePath.Cleaned(), "'", "'\\''")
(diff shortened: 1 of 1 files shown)


MrSerth:
CodeOcean has been updated as well: https://github.com/openHPI/codeocean/commit/8e403ff65ab9da40996dffa56b1a3c44e1e8b2b7

PR #168, opened 2022-06-16
#166 Adjust outdated make command pattern

### LINKED ISSUE #166: AWS: Make not found
All users on AWS are experiencing issues with CodeOcean. For them, the execution is no longer working as `make` cannot be found:

```
sh: make: command not found

The last code run exited with a failure (status code 127).
```

As seen in the [AWS exercise](https://codeocean.openhpi.de/exercises/1005), this is bound to the AWS method and probably started on Sunday, June 12th.
@mpass99 Can you have a look, please?

mpass99:
as with #147 a new unset command part is introduced.

Also add a regression e2e test that would have caught this issue.

Closes #166

### DIFF
--- a/deploy/aws/java11Exec/src/main/java/poseidon/App.java
+++ b/deploy/aws/java11Exec/src/main/java/poseidon/App.java
@@ -67,9 +67,10 @@ public class App implements RequestHandler<APIGatewayV2WebSocketEvent, APIGatewa
         return cmd[cmd.length - 1];
     }
 
-    // Wrapps the passed command with "sh -c".
-    public static String[] wrapCommand(String cmd) {
-        return new String[]{"sh", "-c", cmd};
+    // Replaces the last element of the command with the new shell command.
+    public static String[] wrapCommand(String[] originalCommand, String shellCommand) {
+        originalCommand[originalCommand.length - 1] = shellCommand;
+        return originalCommand;
     }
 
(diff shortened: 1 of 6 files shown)


MrSerth (commented):
Looks good so far; should we also add an assertion about the (removed) AWS environment variables?

mpass99:
We have already made assertions about the environment variables in another test:

https://github.com/openHPI/poseidon/blob/79dc3a94daf62f3b6301246fe20935e916a5553c/tests/e2e/websocket_test.go#L176-L180

Would you still add another assertion here or is it fine with them?

MrSerth:
> We have already made assertions about the environment variables in another test:
> 
> https://github.com/openHPI/poseidon/blob/79dc3a94daf62f3b6301246fe20935e916a5553c/tests/e2e/websocket_test.go#L176-L180
> 
> Would you still add another assertion here or is it fine with them?

Thanks for pointing that out; I also thought of the existing test. Actually, I thought of adding it here as well (just to ensure that the interplay of make and the env variables work as expected) but on second thought, this might also be too much for this test case. I could also think of another one having that or would agree if you convince me that this is not needed (as the combination should also be fine if both separate tests work).

mpass99:
> to ensure that the interplay of make and the env variables work as expected

Good catch! It did not.. With our past rexex we were just ignoring the parts of the command before and after the `make` call... I hope with the next AWS deployment (after you have a chance to take a quick look over the changes) it will work

MrSerth (approved):
Looks good, thanks for including the refactoring!

PR #169, opened 2022-06-19
#155 Minimise timing issues with websocket close.

mpass99:
In the first commit we moved the condition check of the internal status of the WebSocket connection.
When closing the WebSocket connection after an execution...
https://github.com/openHPI/poseidon/blob/79dc3a94daf62f3b6301246fe20935e916a5553c/internal/api/websocket.go#L268
https://github.com/openHPI/poseidon/blob/79dc3a94daf62f3b6301246fe20935e916a5553c/internal/api/websocket.go#L292-L295

...we send the closing message after we closed the WebSocket connection.

But, when we first send the close message
https://github.com/openHPI/poseidon/blob/79dc3a94daf62f3b6301246fe20935e916a5553c/internal/api/websocket.go#L300

before stopping the WebSocket read internally. We receive errors on the tcp level. Therefore we had to introduce the option to stop the WebSocket reading separately before closing also the WebSocket writing.

### DIFF
--- a/internal/api/websocket.go
+++ b/internal/api/websocket.go
@@ -33,6 +33,7 @@ type WebSocketReader interface {
 	io.Reader
 	io.Writer
 	startReadInputLoop()
+	stopReadInputLoop()
 }
 
 // codeOceanToRawReader is an io.Reader implementation that provides the content of the WebSocket connection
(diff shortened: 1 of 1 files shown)


MrSerth (approved):
That's a tough one, but I think, I got it. By introducing a separate `cancelReadCtx`, we are able to separate reading from the input loop and actually closing the WebSocket connection.

I would suggest to ship it and observe the new behaviour; maybe it's enough to reduce the error. Otherwise, I think we might need further investigation to remove that timing issue.

PR #183, opened 2022-07-31
Storage - Delete: Catch non existent removals

mpass99:
Depends on #182 

Another solution would be to avoid calling the "Delete" function multiple times. We do so [here](https://github.com/openHPI/poseidon/blob/7edd40b4f0b55d44fa2f05293098191e06b91ac2/internal/runner/nomad_manager.go#L72), [here](https://github.com/openHPI/poseidon/blob/7edd40b4f0b55d44fa2f05293098191e06b91ac2/internal/runner/nomad_manager.go#L176) and [here](https://github.com/openHPI/poseidon/blob/7edd40b4f0b55d44fa2f05293098191e06b91ac2/internal/runner/nomad_manager.go#L179).

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -11,7 +11,7 @@ require (
 	github.com/hashicorp/nomad/api v0.0.0-20220607124903-78553f9aad20
 	github.com/influxdata/influxdb-client-go/v2 v2.9.1
 	github.com/mitchellh/mapstructure v1.5.0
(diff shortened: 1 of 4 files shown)


MrSerth:
Fixing the call seems to be the first solution. We could, of course, also ensure that we don't call the method multiple times, but this looks also good :)

PR #186, opened 2022-08-02
Update dependencies in Java POM

MrSerth:
A small dependency update for Java


MrSerth:
I updated everything I could think of and also modified all Makefiles in our production environment to the new version.
@mpass99 Please have a look :)

mpass99 (approved):

PR #188, opened 2022-08-03
Fixes for the Grafana Dashboard

mpass99:
**Monitor environment id also for deletions.**
Fixes "Check why the idle runner count always equals the prewarming pool size"

Since only **new** runners have the environment id, the others are filtered out. The **removed** runners contain the lower counts for the idle runners, these values are currently not displayed.

**Fix startup time format**
Currently the startup duration is not efficiently usable since it is stored as string with different types (second, milliseconds,..). We want the duration to be uniform.
We may need to delete all existing startup duration records as flux does not allow differing types.

### DIFF
--- a/internal/runner/abstract_manager.go
+++ b/internal/runner/abstract_manager.go
@@ -30,8 +30,8 @@ func NewAbstractManager() *AbstractManager {
 }
 
 // MonitorRunnersEnvironmentID passes the id of the environment e into the monitoring Point p.
-func MonitorRunnersEnvironmentID(p *write.Point, e Runner, isDeletion bool) {
-	if !isDeletion && e != nil {
+func MonitorRunnersEnvironmentID(p *write.Point, e Runner, _ bool) {
+	if e != nil {
 		p.AddTag(monitoring.InfluxKeyEnvironmentID, e.Environment().ToString())
 	}
 }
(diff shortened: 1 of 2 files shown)


MrSerth:
Thanks for your detailed description, which made reviewing much easier -- I appreciate that! Also, I am fine if you delete the incorrect data in our existing data :)

MrSerth (approved):

mpass99:
> Fix startup time format

According to [the official FAQ](https://docs.influxdata.com/influxdb/v2.3/reference/faq/#can-i-change-a-fields-data-type) we cannot change the type of the field afterwards.
I have tried many workarounds but they do not seem to work (with our version). We could:
- Continue saving the duration as string and convert it on query-time via [flux](https://docs.influxdata.com/flux/v0.x/stdlib/universe/duration/#convert-a-string-to-a-duration).
- Change the fields name. Instead of `duration`, `startup_duration` for example.
- Copy the data with the new field type into a new bucket: `poseidon_v2`

Which solution would you prefer @MrSerth ?

MrSerth:
I would proceed with a clean way and actually change the field (even though this results in some lost data). 

While the last suggestion to copy everything to a new bucket seems to be a bit too much, I would go with the second solution (renaming the field) and potentially drop the outdated filed and data once we finished our analysis for the paper.

PR #197, opened 2022-08-11
#110 Add Grafana dashboard visualizing important Poseidon stats.

### LINKED ISSUE #110: Create Grafana Dashboards

mpass99:
Closes #110 

The open ToDo's of #110 seem to be implemented. What do you think?

### DIFF
--- a/pkg/storage/storage.go
+++ b/pkg/storage/storage.go
@@ -1,12 +1,13 @@
 package storage
 
 import (
+	"context"
 	influxdb2 "github.com/influxdata/influxdb-client-go/v2"
 	"github.com/influxdata/influxdb-client-go/v2/api/write"
 	"github.com/openHPI/poseidon/pkg/logging"
 	"github.com/openHPI/poseidon/pkg/monitoring"
-	"strconv"
 	"sync"
+	"time"
 )
 
 var log = logging.GetLogger("storage")
@@ -43,7 +44,18 @@ type Storage[T any] interface {
 	Sample() (o T, ok bool)
 }
 
-type WriteCallback[T any] func(p *write.Point, object T, isDeletion bool)
+// EventType is an enum type to declare the different causes of a monitoring event.
+type EventType string
+
+const (
+	Creation     EventType = "creation"
+	Deletion     EventType = "deletion"
+	Periodically EventType = "periodically"
+)
+
+// WriteCallback is called before an event gets monitored.
+// Iff eventType is Periodically it is no object provided.
+type WriteCallback[T any] func(p *write.Point, object T, eventType EventType)
 
 // localStorage stores objects in the local application memory.
 type localStorage[T any] struct {
@@ -64,12 +76,18 @@ func NewLocalStorage[T any]() *localStorage[T] {
 // NewMonitoredLocalStorage responds with a Storage implementation.
 // All write operations are monitored in the passed measurement.
 // Iff callback is set, it will be called on a write operation.
-func NewMonitoredLocalStorage[T any](measurement string, callback WriteCallback[T]) *localStorage[T] {
-	return &localStorage[T]{
+// Iff additionalEvents not zero, the duration will be used to periodically send additional monitoring events.
+func NewMonitoredLocalStorage[T any](
+	measurement string, callback WriteCallback[T], additionalEvents time.Duration) *localStorage[T] {
+	s := &localStorage[T]{
 		objects:     make(map[string]T),
 		measurement: measurement,
 		callback:    callback,
 	}
+	if additionalEvents != 0 {
+		go s.periodicallySendMonitoringData(additionalEvents)
+	}
+	return s
 }
 
 func (s *localStorage[T]) List() (o []T) {
@@ -85,7 +103,7 @@ func (s *localStorage[T]) Add(id string, o T) {
 	s.Lock()
 	defer s.Unlock()
 	s.objects[id] = o
-	s.sendMonitoringData(id, o, false, s.unsafeLength())
+	s.sendMonitoringData(id, o, Creation, s.unsafeLength())
 }
 
 func (s *localStorage[T]) Get(id string) (o T, ok bool) {
@@ -101,7 +119,7 @@ func (s *localStorage[T]) Delete(id string) {
 	o, ok := s.objects[id]
 	if ok {
 		delete(s.objects, id)
-		s.sendMonitoringData(id, o, true, s.unsafeLength())
+		s.sendMonitoringData(id, o, Deletion, s.unsafeLength())
 	}
 }
 
@@ -115,7 +133,7 @@ func (s *localStorage[T]) Purge() {
 	s.Lock()
 	defer s.Unlock()
 	for key, object := range s.objects {
-		s.sendMonitoringData(key, object, true, 0)
+		s.sendMonitoringData(key, object, Deletion, 0)
 	}
 	s.objects = make(map[string]T)
 }
@@ -125,7 +143,7 @@ func (s *localStorage[T]) Sample() (o T, ok bool) {
 	defer s.Unlock()
 	for key, object := range s.objects {
 		delete(s.objects, key)
-		s.sendMonitoringData(key, object, true, s.unsafeLength())
+		s.sendMonitoringData(key, object, Deletion, s.unsafeLength())
 		return object, true
 	}
 	return o, false
@@ -143,17 +161,26 @@ func (s *localStorage[T]) unsafeLength() uint {
 	return uint(length)
 }
 
-func (s *localStorage[T]) sendMonitoringData(id string, o T, isDeletion bool, count uint) {
+func (s *localStorage[T]) sendMonitoringData(id string, o T, eventType EventType, count uint) {
 	if s.measurement != "" {
 		p := influxdb2.NewPointWithMeasurement(s.measurement)
 		p.AddTag("id", id)
-		p.AddTag("isDeletion", strconv.FormatBool(isDeletion))
+		p.AddTag("event_type", string(eventType))
 		p.AddField("count", count)
 
 		if s.callback != nil {
-			s.callback(p, o, isDeletion)
+			s.callback(p, o, eventType)
 		}
 
 		monitoring.WriteInfluxPoint(p)
+	}
+}
+
+func (s *localStorage[T]) periodicallySendMonitoringData(d time.Duration) {
+	ctx := context.Background()
+	for ctx.Err() == nil {
+		stub := new(T)
+		s.sendMonitoringData("", *stub, Periodically, s.Length())
+		<-time.After(d)
 	}
 }
--- a/deploy/grafana-dashboard/main.json
+++ b/deploy/grafana-dashboard/main.json
@@ -3,7 +3,10 @@
     "list": [
       {
         "builtIn": 1,
-        "datasource": "-- Grafana --",
+        "datasource": {
+          "type": "datasource",
+          "uid": "grafana"
+        },
         "enable": true,
         "hide": true,
         "iconColor": "rgba(0, 211, 255, 1)",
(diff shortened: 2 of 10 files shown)


MrSerth:
I like the dashboard, it is really beautiful 🧑‍🎨 

I just have a few remaining questions (checked with a time range of 24h)

- For the idle Runner line chart, hovering with my courses yields 19.5 runners. How could that be or is that just a visualisation issue?
- Could we ensure that the coolers per execution environment are unique and equal across all boards? I asked that previously, I know. Today, the "Executions per runner" shows 10 and 28 in green, whereas the "Executions per minute" had yellow for 28 and the "Execution duration" used blue for 28. 
- Can we fix the color for the two new gauges ("current environment count" and "currently used runners")? As the one with the most has a lighter color, staging had a gray color on the left and production a gray on the right. Not that important, though.
- My all-time-favoritue: What about a mapping for the id to a user-friendly name? 🤓

mpass99:
> * For the idle Runner line chart, hovering with my courses yields 19.5 runners. How could that be or is that just a visualisation issue?

This is because of `aggregateWindow(every: v.windowPeriod, fn: mean, createEmpty: false)`. If you zoom into a smaller timespan the 19.5 values resolve to multiple events having a value of 20 or 19. We could change it to 
```flux
  |> aggregateWindow(
    every: v.windowPeriod,
    fn: (tables=<-, column) => tables |> median(method: "exact_selector"),
    createEmpty: false
  )
```
to avoid the fractional values.

I double checked this with the following query resulting in no data:
```flux
import "math"
from(bucket: "poseidon/autogen")
  |> range(start: v.timeRangeStart, stop: v.timeRangeStop)
  |> filter(fn: (r) => r["_field"] == "count")
  |> filter(fn: (r) => math.modf(f: float(v: r._value)).frac != 0)
```

MrSerth:
Thanks for checking, your analysis sounds valid and reassures me that everything is fine. I also checked your suggestion, and would prefer something (such as a median) to ensure we are not showing any fraction values here. Besides changing that, could we also change the number of "max data points" (in the query options)? At least for a timespan larger than 6 hours (12 hours and more), the default seems a bit too small and hides some details (i.e. ~1300 on my machine in single view mode, which is not that much for 24 hours).

mpass99:
Hm, I like that the "max data points" value is set dynamically to the width of the panel. Because of that I would rather not change it to a static value. But since the value in combination with the time range generates the `windowPeriod` I have adjusted this value this way (increased the data points by factor 5):
```flux
myWindowPeriod = if int(v: v.windowPeriod) >= int(v: 30s) then duration(v: int(v: v.windowPeriod) * 5) else v.windowPeriod
```

mpass99:
> * Could we ensure that the coolers per execution environment are unique and equal across all boards?

Unfortunately I have not found such a functionality within Grafana. Instead, I remembered that in the context of Grafana as Code including Grafanalib and jsonnet this is claimed as one of their advantages.
Because I was interested and had not much motivation to copy paste the color overwrites to the 12 panels that should be synced, I granted myself a timebox to setup grafanalib. You can see the results of this timebox in the WIP commit (that I can revert if we conclude so).

I find it useful. It is just writing the dashboard json in a more structured way with the potential of reusing parts such as our color mapping. It takes effort to translate our dashboard into python but this is way less than the effort for creating the panels and queries. The pros and cons collected [here](https://github.com/openHPI/poseidon/issues/110#issuecomment-1102349869) are still valid. My biggest issue was splitting the Python script into several files (which I still couldn't figure out).

mpass99:
> * Can we fix the color for the two new gauges ("current environment count" and "currently used runners")?

Maybe the gauge type was not that fitting. I have changed it into a "Stat".

mpass99:
> * My all-time-favoritue: What about a mapping for the id to a user-friendly name? :nerd_face:

As discussed I have done this by joining with the `poseidon_environments` measurement.

MrSerth:
> Hm, I like that the "max data points" value is set dynamically to the width of the panel. Because of that I would rather not change it to a static value. But since the value in combination with the time range generates the `windowPeriod` I have adjusted this value this way (increased the data points by factor 5):
> 
> ```
> myWindowPeriod = if int(v: v.windowPeriod) >= int(v: 30s) then duration(v: int(v: v.windowPeriod) * 5) else v.windowPeriod
> ```

I also agree and prefer the "max data points" value to be set dynamically. Maybe I am just surprised by the visualisation, as it does not align with my expectation: 
- For the last days, I would expect that the dashboard mostly shows 20 runners for Python:3.8. That's not the case, but rather a long line of 19 (2022-08-14, 24hours). 
- Also, I would expect to see small changes, e.g., from 20 to 19 if a runner is used. With the current visualisation (incl. your suggestions), those differences are not visible but flatten out. If I dramatically increase the "max data points" to 12.000 points, I see those smaller differences with weird? triangle shapes (90° angle from 19 to 20 and a diagonal line to the next 19 data point).

Does it make sense what I am describing? Maybe, my comment about the "max data points" was not specific enough, and it might not fix my problem, but rather reveal some more questions 😅

MrSerth:
> > * Could we ensure that the coolers per execution environment are unique and equal across all boards?
> 
> Unfortunately I have not found such a functionality within Grafana. Instead, I remembered that in the context of Grafana as Code including Grafanalib and jsonnet this is claimed as one of their advantages. Because I was interested and had not much motivation to copy paste the color overwrites to the 12 panels that should be synced, I granted myself a timebox to setup grafanalib. You can see the results of this timebox in the WIP commit (that I can revert if we conclude so).
> 
> I find it useful. It is just writing the dashboard json in a more structured way with the potential of reusing parts such as our color mapping. It takes effort to translate our dashboard into python but this is way less than the effort for creating the panels and queries. The pros and cons collected [here](https://github.com/openHPI/poseidon/issues/110#issuecomment-1102349869) are still valid. My biggest issue was splitting the Python script into several files (which I still couldn't figure out).

Thanks for taking the time and testing that! In general it looks pretty promising -- I would just be interested how to "deploy" it. I see three options:

- Drop the requirement
- Use the "Grafana as code" approach
- Check wether we could provide this data through Poseidon / CodeOcean to influx. Similar to the environment name, we could add an additional attribute to define a color there. This would also allow us to use the same color on the legacy visualisations in CodeOcean.

What do you think, especially about the third option? If we want to continue with the second, I would propose to extract that to a different ticket.

mpass99:
Good questions! It was very interesting and helpful for me gaining an understanding of why this is so.

> * For the last days, I would expect that the dashboard mostly shows 20 runners for Python:3.8. That's not the case, but rather a long line of 19 (2022-08-14, 24hours).

Often there are exactly as much events having `19` idle runners as there are events with `20` runners. In this case `median(method: "exact_selector")` returns `19`. A side cause may that sometimes there are even less than `19` idle runners - leading the median to be smaller.

> I see those smaller differences with weird? triangle shapes (90° angle from 19 to 20 and a diagonal line to the next 19 data point).

Let's first take a look at the 90° angle. When do these angles appear? When we join the graph of idle runners (green) with the events of new started runner (yellow, static arbitrary value) we see that they happen in the moment someone requests a new runner.
Idle Runner vs startup duration

In the panel `Runner startup duration` we see that for this environment a runner needs about 700ms to startup. If we take a closer look at the 90° we can observe that its not exactly 90° - between those events is about one second. This second is matching with the 700ms for the runner startup. So we see that a runner is used (value `19`) and it is immediately refilled (value `20`).

The "diagonal line to the next 19 data point" is caused by the fact that we do not send events in between - but if we did they would have the value `20`.

mpass99:
> What do you think, especially about the third option? If we want to continue with the second, I would propose to extract that to a different ticket.

Also the third option sounds good but unfortunately I do not know enough about Grafana to change the color based on some value (nor I found information to it). The only option I know about is to use the `overwrites` (as we already do at some places)(Also `thresholds` can change the color). Do you know such a Grafana feature to use a passed value as color?

MrSerth:
> In the panel `Runner startup duration` we see that for this environment a runner needs about 700ms to startup. If we take a closer look at the 90° we can observe that its not exactly 90° - between those events is about one second. This second is matching with the 700ms for the runner startup. So we see that a runner is used (value `19`) and it is immediately refilled (value `20`).

Thanks for explaining and digging into that. What if we send two events each time we use a runner (old value / current value) or start a new runner (before starting / once started)? Based on your previous screenshot, I made a drawing:

Sketch

There, each runner used has four events (instead of the two we have currently). This would allow us to always connect two events with a straight line that would represent the current value at that time. Maybe, we would also need to redefine the usage of the median value in this case, but I feel we don't have the correct / complete data yet.

MrSerth:
> > What do you think, especially about the third option? If we want to continue with the second, I would propose to extract that to a different ticket.
> 
> Also the third option sounds good but unfortunately I do not know enough about Grafana to change the color based on some value (nor I found information to it). The only option I know about is to use the `overwrites` (as we already do at some places and would also do with the second option)(Also `thresholds` can change the color). Do you know such a Grafana feature to use a passed value as color?

Unfortunately, I also don't know enough about Grafana to achieve that. Also my web search did not yield any programmatic way, people only refer to the (manual) overrides. It seems that we cannot use this option or would need to change it per graph. Maybe, the grafana-as-code option is the easier and more sustainable one?

mpass99:
> Thanks for explaining and digging into that. What if we send two events each time we use a runner (old value / current value) or start a new runner (before starting / once started)? Based on your previous screenshot, I made a drawing:
> 
> There, each runner used has four events (instead of the two we have currently). This would allow us to always connect two events with a straight line that would represent the current value at that time. Maybe, we would also need to redefine the usage of the median value in this case, but I feel we don't have the correct / complete data yet.

In our discussion we conclude that we use the Grafana line interpolation "step after" that makes the graph look like in the drawing. Additionally we want to add periodically events (each minute) and use the `min` aggregator instead of the `median`.

mpass99:
After the merge we need to adjust some queries due to the new tag "event_type".

pkg/storage/storage.go
```diff
@@ -143,17 +161,26 @@ func (s *localStorage[T]) unsafeLength() uint {
...
 		monitoring.WriteInfluxPoint(p)
 	}
 }
+
+func (s *localStorage[T]) periodicallySendMonitoringData(d time.Duration) {
+	ctx := context.Background()
+	for ctx.Err() == nil {
+		stub := new(T)
+		s.sendMonitoringData("", *stub, Periodically, s.Length())
+		time.Sleep(d)
```
MrSerth, line 184: Is `sleep` appropriate in Go? In other systems, the usage of sleep is usually discouraged. In many cases, developers need to be careful not to send the current thread to sleep (which would also block other executions). Therefore, many have some reactivation, timer, ... concept. According to my search, Go would also feature a [Ticker](https://pkg.go.dev/time#Ticker) struct respectively [NewTicker](https://pkg.go.dev/time#NewTicker) function.

MrSerth (commented):
I like the addition of the periodical monitoring events and just have a minor remark.

pkg/storage/storage.go
mpass99: In Go we just sleep the current Go routine and this way the scheduler does not block other executions.
I searched for other occurrences of `sleep` in our code and found only few. Instead we often use `<-time.After(...)`. I expect those variants to be similar but if you want we can use this instead?
With the Ticker we would also wait and receive from a channel `<-ticker.C` but with the overhead of Creating the ticker before and stopping it afterwards.

MrSerth: Thanks for digging into this, I was also reading some more. A nice [comparison is on StackOverflow](https://stackoverflow.com/a/64500114), but there is no "clear" result. One improvement described there is the ability to cancel `time.After` immediately, which I think is nice (and more responsive?). If you don't have a preference, I would suggest to use `time.After` (as we have done in the code already, `time.sleep` has so far only been used in tests).

MrSerth (approved):
Looks great, thanks!

PR #198, opened 2022-08-16
Update dependencies and pin two Nomad deps

MrSerth:
This MR will update our dependencies to the newest version (see first commit).
Additionally, we pin two sub-dependencies from Nomad to ease future update processes and resolve a security warning (second commit).


mpass99 (approved):
Nice :+1:

PR #200, opened 2022-08-17
Grafana as code

### LINKED ISSUE #199: "Grafana as Code"
> > > * Could we ensure that the coolers per execution environment are unique and equal across all boards?
> > 
> > 
> > Unfortunately I have not found such a functionality within Grafana. Instead, I remembered that in the context of Grafana as Code including Grafanalib and jsonnet this is claimed as one of their advantages. Because I was interested and had not much motivation to copy paste the color overwrites to the 12 panels that should be synced, I granted myself a timebox to setup grafanalib. You can see the results of this timebox in the WIP commit (that I can revert if we conclude so).
> > I find it useful. It is just writing the dashboard json in a more structured way with the potential of reusing parts such as our color mapping. It takes effort to translate our dashboard into python but this is way less than the effort for creating the panels and queries. The pros and cons collected [here](https://github.com/openHPI/poseidon/issues/110#issuecomment-1102349869) are still valid. My biggest issue was splitting the Python script into several files (which I still couldn't figure out).
> 
> Thanks for taking the time and testing that! In general it looks pretty promising -- I would just be interested how to "deploy" it. I see three options:
> 
>     * Drop the requirement
> 
>     * Use the "Grafana as code" approach
> 
>     * Check wether we could provide this data through Poseidon / CodeOcean to influx. Similar to the environment name, we could add an additional attribute to define a color there. This would also allow us to use the same color on the legacy visualisations in CodeOcean.
> 
> 
> What do you think, especially about the third option? If we want to continue with the second, I would propose to extract that to a different ticket.

> Also the third option sounds good but unfortunately I do not know enough about Grafana to change the color based on some value (nor I found information to it). The only option I know about is to use the overwrites (as we already do at some places and would also do with the second option)(Also thresholds can change the color). Do you know such a Grafana feature to use a passed value as color?

> Unfortunately, I also don't know enough about Grafana to achieve that. Also my web search did not yield any programmatic way, people only refer to the (manual) overrides. It seems that we cannot use this option or would need to change it per graph. Maybe, the grafana-as-code option is the easier and more sustainable one?

See also #197

mpass99:
Closes #199


mpass99:
Open ToDos:
- Color mapping
- flux code reusage (at least the envMapping)
-

deploy/grafana-dashboard/availability_row.py
```diff
@@ -0,0 +1,51 @@
...
+    targets=[InfluxDBTarget(query=read_query("prewarming-pool-size"))],
+    gridPos=GridPos(h=10, w=11, x=0, y=3),
+    allValues=True,
+    orientation=ORIENTATION_VERTICAL,
+    displayMode=GAUGE_DISPLAY_MODE_BASIC,
+    max=None,
+)
+
+idle_runner = TimeSeries(
+    title="Idle Runner",
```
MrSerth, line 19: I was wondering whether we should allow more fine granular data points here (similar to the current dashboard version) to account for smaller differences.

deploy/grafana-dashboard/availability_row.py
```diff
@@ -0,0 +1,51 @@
...
+    title="Availability",
+    collapsed=True,
+    gridPos=GridPos(h=1, w=24, x=0, y=2),
+    panels=[
+        prewarming_pool_size,
+        idle_runner,
+        runner_startup_duration,
+        used_runner
+    ]
+)
```
MrSerth, line 51: 🤓 Feel free to add a newline below ;)

deploy/grafana-dashboard/availability_row.py
```diff
@@ -0,0 +1,51 @@
...
+used_runner = TimeSeries(
+    title="Used Runner",
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("used-runner"))],
+    gridPos=GridPos(h=10, w=12, x=12, y=13),
+)
+
+availability_row = RowPanel(
+    title="Availability",
+    collapsed=True,
```
MrSerth, line 43: Should we generally open all rows? I am unsure...

deploy/grafana-dashboard/general_row.py
```diff
@@ -0,0 +1,112 @@
...
+    title="Requests per minute",
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("requests-per-minute"))],
+    gridPos=GridPos(h=9, w=8, x=0, y=1),
+    scaleDistributionType="log",
+    extraJson=grey_all_mapping
+)
+
+request_latency = Heatmap(
+    title="Request Latency",
```
MrSerth, line 17: I cannot tell why, but there seems to be a difference in the heat map (maybe due to the resolution and number of data points)?

Grafana as code:

Bildschirm­foto 2022-10-27 um 00 01 29

Current dashboard:

Bildschirm­foto 2022-10-27 um 00 01 33

deploy/grafana-dashboard/general_row.py
```diff
@@ -0,0 +1,112 @@
...
+    extraJson={
+        "options": {},
+        "yAxis": {
+            "format": "ns"
+        }
+    }
+)
+
+service_time = TimeSeries(
+    title="Service time (99.9%)",
```
MrSerth, line 31: I was wondering whether we should allow more fine granular data points here (similar to the current dashboard version) to account for smaller differences.

deploy/grafana-dashboard/general_row.py
```diff
@@ -0,0 +1,112 @@
...
+    alignment="center"
+)
+
+number_of_executions = BarGauge(
+    title="Number of Executions",
+    dataSource="Poseidon",
+    targets=[InfluxDBTarget(query=read_query("number-of-executions"))],
+    gridPos=GridPos(h=6, w=8, x=16, y=10),
+    allValues=True,
+    displayMode=GAUGE_DISPLAY_MODE_GRADIENT,
```
MrSerth, line 62: Shall we rotate the bars, similar to our current design?

```suggestion
    orientation=ORIENTATION_VERTICAL,
    displayMode=GAUGE_DISPLAY_MODE_BASIC,
```

deploy/grafana-dashboard/poseidon.dashboard.py
```diff
@@ -0,0 +1,25 @@
...
+    title="Poseidon autogen",
+    timezone="browser",
+    panels=[
+        general_row,
+        runner_insights_row,
+        availability_row
+    ],
+    templating=Templating(list=[
+        stage_variable,
+        environment_variable
```
MrSerth, line 18: Something seems different here to our current dashboard:

- Previously, we also allowed to filter for status code, which only seems useful to some graphs
Bildschirm­foto 2022-10-27 um 00 01 19

- Stage is fixed to production instead of allowing other environments as well?
Bildschirm­foto 2022-10-27 um 00 01 00Bildschirm­foto 2022-10-27 um 00 01 04

- Environment ID is not showing all IDs (similar to the graph "Prewarming Pool Size") and there is a "null" entry
Bildschirm­foto 2022-10-27 um 00 01 08Bildschirm­foto 2022-10-27 um 00 01 12

deploy/grafana-dashboard/runner_insights_row.py
```diff
@@ -0,0 +1,66 @@
...
+
+execution_duration_extra_json = {
+    "fieldConfig": {
+        "defaults": {
+            "unit": "ns"
+        }
+    }
+}
+execution_duration = Histogram(
+    title="Execution duration",
```
MrSerth, line 14: Here, I am also wondering why we see a difference between the old and new dashboard, for example with regard to the scale of the y axis (only ~15s instead of ~30s)

Bildschirm­foto 2022-10-27 um 00 03 06

deploy/grafana-dashboard/runner_insights_row.py
```diff
@@ -0,0 +1,66 @@
...
+    title="Executions per runner",
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("executions-per-runner-hist"))],
+    gridPos=GridPos(h=10, w=11, x=0, y=10),
+    bucketSize=1,
+    colorMode="palette-classic",
+)
+
+executions_per_minute = TimeSeries(
+    title="Executions per minute",
```
MrSerth, line 33: I was wondering whether we should allow more fine granular data points here (similar to the current dashboard version) to account for smaller differences.

deploy/grafana-dashboard/runner_insights_row.py
```diff
@@ -0,0 +1,66 @@
...
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("executions-per-minute-time"))],
+    gridPos=GridPos(h=10, w=13, x=11, y=10),
+)
+
+request_body_size = TimeSeries(
+    title="Request Body Size",
```
MrSerth, line 40: I was wondering whether we should allow more fine granular data points here (similar to the current dashboard version) to account for smaller differences.

deploy/grafana-dashboard/runner_insights_row.py
```diff
@@ -0,0 +1,66 @@
...
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("request-body-size"))],
+    gridPos=GridPos(h=10, w=11, x=0, y=20),
+    scaleDistributionType="log",
+    unit="bytes",
+)
+
+runner_per_minute = TimeSeries(
+    title="Runner per minute",
```
MrSerth, line 49: I was wondering whether we should allow more fine granular data points here (similar to the current dashboard version) to account for smaller differences.

deploy/grafana-dashboard/util.py
```diff
@@ -0,0 +1,5 @@
+
+
```
MrSerth, line 2: Is there any specific reason to have two empty lines here? Just wondering...

deploy/grafana-dashboard/variables.py
```diff
@@ -0,0 +1,25 @@
...
+environment_variable = Template(
+    dataSource="Poseidon",
+    label="Environment IDs",
+    name="environment_ids",
+    query=read_query("environment-ids"),
+    refresh=1,
+    includeAll=True,
+    multi=True,
+    default="$__all"
+)
```
MrSerth, line 25: 🤓 Feel free to add a newline here as well ;)

deploy/grafana-dashboard/availability_row.py
```diff
@@ -0,0 +1,51 @@
+from grafanalib.core import RowPanel, BarGauge, GridPos, TimeSeries, ORIENTATION_VERTICAL, \
+    GAUGE_DISPLAY_MODE_BASIC
+from grafanalib.influxdb import InfluxDBTarget
+
+from util import read_query
+
+prewarming_pool_size = BarGauge(
+    title="Prewarming Pool Size",
```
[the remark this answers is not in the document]
mpass99: Now, the panel shows only environments that have been active in the selected time interval. Before, it showed all environments that have been active in the last year. See changes here:
https://github.com/openHPI/poseidon/blob/b5b1173571b338d458bcb3dd4211570fb9831a9f/deploy/grafana-dashboard/queries/environment-ids.flux#L2

Which behaviour do we prefer? (Sorry, for including logical changes and not just doing a 1-to-1 transfer)

deploy/grafana-dashboard/availability_row.py
mpass99: The difference to the original dashboard comes from grafanalib setting a default value for `maxDataPoints`. Before, we have not set a value and so Grafana determines one automatically based on the panel width.
For now, I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/availability_row.py
```diff
@@ -0,0 +1,51 @@
...
+runner_startup_duration = TimeSeries(
+    title="Runner startup duration",
+    dataSource='Poseidon',
+    targets=[InfluxDBTarget(query=read_query("runner-startup-duration"))],
+    gridPos=GridPos(h=10, w=12, x=0, y=13),
+    unit="ns",
+)
+
+used_runner = TimeSeries(
+    title="Used Runner",
```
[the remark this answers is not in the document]
mpass99: I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/general_row.py
mpass99: I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/runner_insights_row.py
mpass99: I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/runner_insights_row.py
mpass99: I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/availability_row.py
mpass99: I have left them collapsed, because otherwise the panels can not be defined nested in the RowPanels anymore (all panels would be listed directly in the Dashboard definition). I am fine with both..

deploy/grafana-dashboard/general_row.py
mpass99: Yeah, it is because of the static number of data points. I have now changed it to a dynamic number of data points (as it was originally).

deploy/grafana-dashboard/runner_insights_row.py
mpass99: I have adjusted this dashboard to use the dynamic value again.

deploy/grafana-dashboard/poseidon.dashboard.py
mpass99: > * Previously, we also allowed to filter for status code, which only seems useful to some graphs

Exactly, the filter for the status code has increased the complexity but had a very limited usage, so I removed it. Are you fine with that?

> Stage is fixed to production instead of allowing other environments as well?

I had falsely just queried the stages of environments that have been active in the selected time range. Now we display all stages of the last year (as in the original dashboard).

> Environment ID is not showing all IDs (similar to the graph "Prewarming Pool Size") and there is a "null" entry

See https://github.com/openHPI/poseidon/pull/200#discussion_r1008734037

deploy/grafana-dashboard/runner_insights_row.py
mpass99: Like for the other panels, this is a result of the static number of `maxDataPoints`. With less displayed data points, all data points need to get more aggregated and so the mean of them is lower.

I have now changed it to a dynamic number again.

deploy/grafana-dashboard/poseidon.dashboard.py
MrSerth: >> Previously, we also allowed to filter for status code, which only seems useful to some graphs
> 
> Exactly, the filter for the status code has increased the complexity but had a very limited usage, so I removed it. Are you fine with that?

Yes, that's absolutely fine for me 👍  Thanks for taking care!

deploy/grafana-dashboard/availability_row.py
MrSerth: Ah, I didn't know that. Having that collapsed is fine; we might just want to uncollapse them after saving a new version (I find it easer to look at the uncollapsed version, but the code looks better with the nested definitions).

mpass99: I also like the uncollapsed version more and I think with [14c4212](https://github.com/openHPI/poseidon/pull/200/commits/14c4212bfb7dc259866ae774d7197071a467aab1) the code quality is not that much influenced..
PS. I changed the order of the rows because I find the availability panels the most interesting?!

MrSerth: Awesome, I just saw that when rebasing (and testing) my branch. Thanks for having another look and working on this solution; I think it is still clean and nice.

And yes, the changed order might be useful to get a good impression on the first glance 🔍

MrSerth (approved):
Wohoo! 🎉

PR #201, opened 2022-08-19
Add environment id to periodical monitoring events.

mpass99:
Otherwise, the new periodical events have no environment id and would not be valuable for the "idle runner" graph.


MrSerth (approved):
Thanks for noticing and fixing 👍

PR #210, opened 2022-09-07
Fix Nomad event stream is ignoring errors

### LINKED ISSUE #204: Poseidon requires restart to send updated data to Influx
We temporarily had a low availability of Runners. Afterwards, Poseidon does not send the most up-to-date data to Grafana, so that we have a discrepancy between the data returned by the API (for the current prewarming pool size) and the data written to Influx. After restarting Poseidon, the data reported automatically increased and was as expected.

Bildschirmfoto 2022-08-27 um 00 10 09

mpass99:
when an event stream could be established once.

Closes #204 

`events.IsHeartbeat()` is false positive for error events. This way we ignored the `unexpected EOF` error that we receive when Nomad restarts.

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -211,11 +211,10 @@ type nomadAPIEventHandler func(event *nomadApi.Event) (done bool, err error)
 func receiveAndHandleNomadAPIEvents(stream <-chan *nomadApi.Events, handler nomadAPIEventHandler) error {
 	// If original context is canceled, the stream will be closed by Nomad and we exit the for loop.
 	for events := range stream {
-		if events.IsHeartbeat() {
-			continue
-		}
 		if err := events.Err; err != nil {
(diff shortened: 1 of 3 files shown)


MrSerth (approved):
Looks great, thanks! Will your fix also apply to situations where we just update an execution environment without restarting Nomad (as caused by the original #204 issue?)

mpass99:
I was not able to reproduce errors with updating an environment..
We will now get an Sentry event for each Event Stream problem. If it happens again we will hopefully have more information.

MrSerth:
Alright, thanks for checking again. I am feeling lucky today and just deployed the new version, let's observe the differences 👍

MrSerth:
Okay, I was able to reproduce it by first updating the execution environment (e.g. with the network-enabled flag) and then synchronizing the environment. I still have this issue (see environment ID 33 on CodeOcean)

PR #215, opened 2022-09-18
#211 Start Daemons and switch user per execution

### LINKED ISSUE #211: Start Daemons in Docker Image
Obviously, we don't start our containers with the `/sbin/my_init` script, that is provided by our [Phusion base image](https://github.com/phusion/baseimage-docker).

The startup script `/sbin/my_init` needs to be executed as root user and can be further [customized to allow additional daemons to be started](https://github.com/phusion/baseimage-docker#running-scripts-during-container-startup). I tried replacing the `sleep infinity` command in `job.go` with `/sbin/my_init`, which didn't work due to the different users (`root` vs `user`). Do we have another idea on how to execute this task as `root` instead (either through Nomad during the execute call or using some Linux mechanism, such as the setuid bit for the file)?

MrSerth:
_Part of #211 and https://github.com/openHPI/dockerfiles/pull/23, Closes #211 and closes #213, //cc @koelibri_

This PR:
- allows using a local Docker image for Development and Testing
- builds the e2e test image if modified in the current branch
- Switches the user per code execution. Unfortunately, this only works in Nomad so far (as AWS does not yet have custom images) and also imposes the first requirement on Docker images (the availability of the `/sbin/setuser` command). I would be glad to prevent the dependency on the Docker images (and the `/sbin/setuser` command) but didn't find a suitable workaround. Should we add some configuration option for this behaviour?

Open ToDos:
- Find a generalizable solution for `/sbin/setuser` dependency.
- Add AWS support.
- Refactor code to be cleaner.

Assistance for the open points is very appreciated!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -97,7 +97,7 @@ deploy/dockerfiles: ## Clone Dockerfiles repository
 	@git clone git@github.com:$(REPOSITORY_OWNER)/dockerfiles.git deploy/dockerfiles
 
 .PHONY: e2e-test-docker-image
-e2e-test-docker-image: deploy/dockerfiles ## Build Docker image that is pushed to a registry and used in e2e tests
+e2e-test-docker-image: deploy/dockerfiles ## Build Docker image that is used in e2e tests
 	@docker build -t $(E2E_TEST_DOCKER_IMAGE) deploy/dockerfiles/$(E2E_TEST_DOCKER_CONTAINER)/$(E2E_TEST_DOCKER_TAG)
 
 .PHONY: e2e-test
--- a/Makefile
+++ b/Makefile
@@ -3,13 +3,14 @@ REPOSITORY_OWNER = "openHPI"
 PKG := "github.com/$(REPOSITORY_OWNER)/$(PROJECT_NAME)/cmd/$(PROJECT_NAME)"
 UNIT_TESTS = $(shell go list ./... | grep -v /e2e)
 
-DOCKER_E2E_CONTAINER_NAME := "$(PROJECT_NAME)-e2e-tests"
 DOCKER_TAG := "poseidon:latest"
 DOCKER_OPTS := -v $(shell pwd)/configuration.yaml:/configuration.yaml
 LOWER_REPOSITORY_OWNER = $(shell echo $(REPOSITORY_OWNER) | tr A-Z a-z)
-REGISTRY = "ghcr.io"
-TAG = ":main"
-E2E_TEST_DOCKER_IMAGE = "$(REGISTRY)/$(LOWER_REPOSITORY_OWNER)/$(DOCKER_E2E_CONTAINER_NAME)$(TAG)"
+
+# Define image to be used in e2e tests. Requires `make` to be available.
+E2E_TEST_DOCKER_CONTAINER := co_execenv_java
+E2E_TEST_DOCKER_TAG := 17
+E2E_TEST_DOCKER_IMAGE = "$(LOWER_REPOSITORY_OWNER)/$(E2E_TEST_DOCKER_CONTAINER):$(E2E_TEST_DOCKER_TAG)"
 
 default: help
 
@@ -92,24 +93,27 @@ coverage: deps ## Generate code coverage report
 coverhtml: coverage ## Generate HTML coverage report
 	@go tool cover -html=coverage_cleaned.cov -o coverage_unit.html
 
-.PHONY: e2e-test-docker-image ## Build Docker image that is pushed to a registry and used in e2e tests
-e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
-	@docker build -t $(E2E_TEST_DOCKER_IMAGE) deploy/e2e-test-image
+deploy/dockerfiles: ## Clone Dockerfiles repository
+	@git clone git@github.com:$(REPOSITORY_OWNER)/dockerfiles.git deploy/dockerfiles
+
+.PHONY: e2e-test-docker-image
+e2e-test-docker-image: deploy/dockerfiles ## Build Docker image that is pushed to a registry and used in e2e tests
+	@docker build -t $(E2E_TEST_DOCKER_IMAGE) deploy/dockerfiles/$(E2E_TEST_DOCKER_CONTAINER)/$(E2E_TEST_DOCKER_TAG)
 
 .PHONY: e2e-test
 e2e-test: deps ## Run e2e tests
-	@docker pull $(E2E_TEST_DOCKER_IMAGE)
+	@[ -z "$(docker images -q $(E2E_TEST_DOCKER_IMAGE))" ] || docker pull $(E2E_TEST_DOCKER_IMAGE)
 	@go test -count=1 ./tests/e2e -v -args -dockerImage="$(E2E_TEST_DOCKER_IMAGE)"
 
 .PHONY: e2e-docker
 e2e-docker: docker ## Run e2e tests against the Docker container
 	docker run --rm -p 127.0.0.1:7200:7200 \
-       --name $(DOCKER_E2E_CONTAINER_NAME) \
+       --name $(E2E_TEST_DOCKER_CONTAINER) \
        -e POSEIDON_SERVER_ADDRESS=0.0.0.0 \
        $(DOCKER_OPTS) \
        $(DOCKER_TAG) &
 	@timeout 30s bash -c "until curl -s -o /dev/null http://127.0.0.1:7200/; do sleep 0.1; done"
-	@make e2e-test || EXIT=$$?; docker stop $(DOCKER_E2E_CONTAINER_NAME); exit $$EXIT
+	@make e2e-test || EXIT=$$?; docker stop $(E2E_TEST_DOCKER_CONTAINER); exit $$EXIT
 
 # See https://aquasecurity.github.io/trivy/v0.18.1/integrations/gitlab-ci/
 TRIVY_VERSION = $(shell wget -qO - "https://api.github.com/repos/aquasecurity/trivy/releases/latest" | grep '"tag_name":' | sed -E 's/.*"v([^"]+)".*/\1/')
--- a/internal/nomad/nomad_test.go
+++ b/internal/nomad/nomad_test.go
@@ -676,28 +676,31 @@ func (s *ExecuteCommandTestSuite) TestWithSeparateStderr() {
 	var calledStdoutCommand, calledStderrCommand []string
 
 	// mock regular call
-	s.mockExecute(s.testCommandArray, commandExitCode, nil, func(args mock.Arguments) {
+	call := s.mockExecute(mock.AnythingOfType("[]string"), 0, nil, func(_ mock.Arguments) {})
+	call.Run(func(args mock.Arguments) {
 		var ok bool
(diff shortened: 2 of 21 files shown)


deploy/e2e-test-image/setuser.py
```diff
@@ -0,0 +1,61 @@
+#!/usr/bin/python3
```
mpass99, line 1: Why do we make an extra Docker image for our e2e tests?

Ok, it is minimal and so more time and energy efficient.
On the other hand, it increases the complexity, requires to add an foreign, static, not auto updated script to our repository and prevents us from using a Docker image that is used in production and not CI/CD tested at this moment.
We could set `E2E_TEST_DOCKER_IMAGE` in the Makefile statically to [`openhpi/co_execenv_python:3.8`](https://github.com/openHPI/dockerfiles/tree/master/co_execenv_python/3.8).

What do you think?

mpass99 (commented):
Great, thanks for this much work! :brain:  

---
> Should we add some configuration option for this behaviour?

Yeah, I would prefer a configuration for the username.

> Open ToDos:

Are they meant as separate issues or as things that need to be done in this PR?

---
Maybe I would add a test such as this one from job_test.go l. 100:

``` Go
func TestSetForcePullFlag(t *testing.T) {
	_, job := helpers.CreateTemplateJob()
	taskGroup := FindAndValidateDefaultTaskGroup(job)
	task := FindAndValidateDefaultTask(taskGroup)

	t.Run("Ignoring passed value if DisableForcePull", func(t *testing.T) {
		config.Config.Nomad.DisableForcePull = true
		SetForcePullFlag(job, true)
		require.Equal(t, false, task.Config["force_pull"])
	})

	t.Run("Using passed value if not DisableForcePull", func(t *testing.T) {
		config.Config.Nomad.DisableForcePull = false
		SetForcePullFlag(job, true)
		require.Equal(t, true, task.Config["force_pull"])

		SetForcePullFlag(job, false)
		require.Equal(t, false, task.Config["force_pull"])
	})
}
```

deploy/e2e-test-image/setuser.py
MrSerth: I was wondering the same, to be honest, and thought we switched to the "famous" `[openhpi/co_execenv_python:3.8](https://github.com/openHPI/dockerfiles/tree/master/co_execenv_python/3.8)` image some time ago. Therefore, I was surprised that we didn't.

For a moment, I also considered switching (especially after configuring everything took me some while) to have the same advantages you mentioned. Sure, having a minimal image is also nice, but I would assume the time and energy benefits are negligible.

My main reason to keep it was to have an independent development cycle for the Docker images and Poseidon. In the current situation, using the "released" version of our Dockerfiles would prevent the successful test of not being root or cause many e2e failures (the `/sbin/setuser` script can only be executed by root). 
An alternative to adding the Python script included in our Images by default would be to change the `FROM` directive in the Dockerfile: Instead of starting with a [Python-Image](https://github.com/openHPI/poseidon/blob/01e5ed0d4e2b6fec9c7738426d4f9c4e9b4305fd/deploy/e2e-test-image/Dockerfile#L2) and adding the `/sbin/setuser` script, we could bring it closer to our production environment:

```dockerfile
FROM phusion/baseimage:master
```

This would allow us to remove the "arbitrary" Python script from the folder, while still keeping an independent Docker file for testing purposes in this repository. However, by bringing that closer, we also reduce the argument not to use the production environment...

What would you prefer?

MrSerth:
Thanks for reviewing! 🙏

> > Should we add some configuration option for this behaviour?
>
> Yeah, I would prefer a configuration for the username.

Just of the username? I thought of adding a config option via our RESTful API to enable and disable switching with this script per execution. In my opinion, it could be similar to the "enable network" functionality that we have.

Would that make sense for you or do you see another solution?

> > Open ToDos:
>
> Are they meant as separate issues or as things that need to be done in this PR?

It depends 😁
If we add some configuration for the `/sbin/setuser`, these could (potentially) resolve automatically (just enable the option for all Nomad-related executions and disable it -- for now -- for AWS). Hopefully, this also allows us to find a good architectural structure and thereby resolve all issues at once.

deploy/e2e-test-image/setuser.py
mpass99: I see the argument of having an independent development cycle. But this means, that we have to do the changes to the Dockerfiles twice - once in the other repo and once here. If we have to do this anyways we could also:

1. Have `FROM openhpi/co_execenv_python:3.8` as default.
2. Going down the base-image-chain as far as required for development and copy the changes to this repo.
3. Merge the [Dockerfiles Repo](https://github.com/openHPI/dockerfiles) first.
4. Return to our default (`FROM openhpi/co_execenv_python:3.8`) here and merge this.

I assume this will not be required often. Maybe it's too complicated? I'm also fine with your solution.

mpass99:
> I thought of adding a config option via our RESTful API to enable and disable switching with this script per execution. In my opinion, it could be similar to the "enable network" functionality that we have.

Sounds even better :)

deploy/e2e-test-image/setuser.py
MrSerth: I updated the GitHub action workflow to checkout a *matching* branch name if possible. 
- With a non-matching branch, the published Docker image is used (see example run with branch "test" [here](https://github.com/openHPI/poseidon/actions/runs/3100012999))
- With a matching brach, the not-yet-merged Dockerfile is used to build a new image (see example run with branch "start_init_scripts [here](https://github.com/openHPI/poseidon/actions/runs/3100015284))

Two additional things to note:
- I switched the Python:3.8 image in favour of Java:17, as our tests require the image to contain the `make` command which is not available in the Python:3.8 image
- The remote branch in the docker files repository has a different name. Unfortunately, I cannot rename that branch remotely without closing my existing MR there. Hence, I duplicated the branch to match the name of this branch and will take care of deleting the duplicated branch once we merged everything.

Makefile
```diff
@@ -93,23 +94,23 @@ coverhtml: coverage ## Generate HTML coverage report
 	@go tool cover -html=coverage_cleaned.cov -o coverage_unit.html
 
 .PHONY: e2e-test-docker-image ## Build Docker image that is pushed to a registry and used in e2e tests
-e2e-test-docker-image: deploy/e2e-test-image/Dockerfile
-	@docker build -t $(E2E_TEST_DOCKER_IMAGE) deploy/e2e-test-image
+e2e-test-docker-image: deploy/dockerfiles/${E2E_TEST_DOCKER_CONTAINER}/$(E2E_TEST_DOCKER_TAG)/Dockerfile
```
mpass99, line 97: Should we add a rule (such as git clone) for `deploy/dockerfiles/${E2E_TEST_DOCKER_CONTAINER}/$(E2E_TEST_DOCKER_TAG)/Dockerfile:` to make this Makefile work even outside the CI/CD?

mpass99: Great :blush:

mpass99 (dismissed):

internal/nomad/nomad_test.go
```diff
@@ -666,6 +666,7 @@ func (s *ExecuteCommandTestSuite) SetupTest() {
 }
 
 const withTTY = true
+const unprivileged = false
```
MrSerth, line 669: Isn't this the opposite meaning of the `privilegedExecution` flag? The value of `unprivileged` is passed to the `ExecuteCommand` function and by reading, I would assume it behaves differently. 

Maybe I just mixing things? Otherwise, I would propose to rename this variable to `privileged` or something similar.

MrSerth (commented):
Awesome, thanks so much for taking over and making this so beautiful! This yields a much better maintainability than it had before 👏

I just have a minor question and will also start working on the CodeOcean part 😉

MrSerth:
I just pushed a minor update to the documentation ✍️

MrSerth (commented):
Awesome, thanks! I consider this PR to be ready-to-go (but cannot approve it because I opened it initially).

internal/nomad/nomad_test.go
mpass99: I also was unsure about this naming :sweat_smile: 
My argumentation was that for lines like this:
https://github.com/openHPI/poseidon/blob/537030e98d88b19e419aa657e438e8ed79add822/internal/nomad/nomad_test.go#L703-L704
it's easier to read without (having to know /) knowing the value of the constant (as the name is explaining it).

But for line 669 it's the opposite and one may think it's privileged.. I have changed the name to `privileged` :ok_hand:

mpass99 (approved):

internal/nomad/nomad_test.go
MrSerth: Ah, I see; a probably we might also have for the `withTTY`, have we? I am not aware of a super-nice naming pattern; the only alternative I am aware of are constants (even when just representing a boolean value). Those could indicate to the reader (without knowing the value) what they are meant for. In this case, it would require us to use two constants: One for the privileged execution and one for the unprivileged execution... 
When having a "normal" variable, I liked having the same semantic meaning of the variable name (regardless of content) with the parameter name (just to ease debugging).

mpass99: Like this?

MrSerth: Yes, like this :) What do you think about it? Does it combine the better readability in the tests without clashing the argument and parameter names?

mpass99: Yes, I like it too :)

mpass99 (approved):

internal/nomad/nomad_test.go
MrSerth: Nice, thanks for taking care. Let's merge it and bring it live(?) 🎉

internal/runner/nomad_runner.go
```diff
@@ -130,6 +130,7 @@ func (r *NomadJob) UpdateFileSystem(copyRequest *dto.UpdateFileSystemRequest) er
 	stdOut := bytes.Buffer{}
 	stdErr := bytes.Buffer{}
 	exitCode, err := r.api.ExecuteCommand(r.id, context.Background(), updateFileCommand, false,
+		nomad.PrivilegedExecution, // All files should be written and owned by a privileged user #211.
```
mpass99, line 133: Is this not an oversimplification? While test files are not supposed to be writable (and readable?), one can well imagine a task in which a given CSV file is to be extended by entries... This distinction is currently not possible. @MrSerth Do we want to extend our API for the UpdateFileRoute by an Owner (, Group and Permissions)?

MrSerth: Hm, good point. 

- Test files need to be readable by `user`, as this user is supposed to execute them. Hence, we cannot remove that flag.
- With regard to the example you provided: Yes, the current implementation is a simplification. However, I am currently not aware of a use case where appending is required, but this might be valid. Instead of adding an explicit owner, group and permissions, I would go a soft way: What if we add one additional boolean flag (besides `path` and `content`) to the UpdateFileRoute indicating whether the user used for execution should have write permissions? Or which permissions (`rwx`) the user should haven general? I would just prevent providing an explicit user name, as this requires further configuration. If set, this could be apply to the "other" group and will therefore work for the `user`. If privilegedExecution is enabled, the "other" group won't be respected, but `root` may always read :)

PR #223, opened 2022-09-30
Specify requirements for Docker images

### LINKED ISSUE #3: Specify requirements on Docker images
As a **user**,  
I want to know the requirements on the Docker images Poseidon can operate on,  
in order to avoid failures due to missing commands.

Currently, file copying inside the container is done using `tar --extract --absolute-names --verbose --file=/dev/stdin`. However, in Alpine containers, for example, the `tar` command does not know the `--absolute-names` option.

There might be other places in the code where implicit assumptions on the functionality available inside the Docker container are made. Those should be specified.

MrSerth:
Closes #3

The list is based on the first ideation we started during our meeting and some code searches. Please have a look whether everything is correct or whether we need to add more commands.


mpass99 (approved):
The code does not use `whoami`  by now but we would like to require it at least for manual testing purposes, right?

mpass99:
Ah, I see, thanks :+1:

PR #229, opened 2022-10-06
Compress files between Nomad and Poseidon

### LINKED ISSUE #227: Investigate compression for file transfer
We might want to investigate a compression mechanism for file transfers:

- File copy: Should we use a compression instead of just bundling with tar?
- Transfer back: Consider piping the `cat` output through a compression program in the container and decompress it in Poseidon to allow for faster data transfer? Is that possible with streaming? The goal would be to reduce the bytes sent through Nomad, as this lead to a very high load for Nomad with large files (I tested it 1G):

  ```ruby
  system("dd if=/dev/zero of=test.txt bs=1 count=0 seek=1G") # Can be compressed very well!
  system("dd if=/dev/random of=test.txt bs=1 count=0 seek=1G") # Cannot be compressed probably
  ```

mpass99:
Closes #227

### DIFF
--- a/internal/runner/nomad_runner.go
+++ b/internal/runner/nomad_runner.go
@@ -3,6 +3,7 @@ package runner
 import (
 	"archive/tar"
 	"bytes"
+	"compress/gzip"
 	"context"
 	"encoding/base64"
 	"encoding/json"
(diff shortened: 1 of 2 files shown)


MrSerth:
As described in the ticket #227, we rather see this PR as a spike showing the feasibility. For now, we won't continue with this approach, but the issue is still valid and could be implemented in the future. Meanwhile, I am closing this PR.

PR #230, opened 2022-10-06
OpenAPI Add token and version prefix.

### LINKED ISSUE #209: OpenAPI Definition - Add Token Header and Version Prefix to Path
In our swagger definition we specify for many routes the `401 Unauthorized` response. Having only this document, this raises the question of how to authenticate.

This may be the appropriate place to add the information about our `Poseidon-Token`:
```openAPI
          parameters:
            - in: header
              name: Poseidon-Token
              schema:
                type: string
                format: password
              required: true
```
https://swagger.io/docs/specification/describing-parameters/#header-parameters

mpass99:
Closes #209

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -5,6 +5,9 @@ info:
     This API is used by CodeOcean to run code in runners.
   version: '0.2.2'
 
+servers:
+  - url: '/api/v1'
+
 components:
   schemas:
     ExecutionEnvironment:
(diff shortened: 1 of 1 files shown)


MrSerth:
Looks good so far, but I just had a quick look. I was wondering about the token parameter; that (of course) is valid. However, with the `security` directive, we could also tell that this header is used to authenticate, couldn't we? [Here is an example](https://swagger.io/docs/specification/authentication/api-keys/).

mpass99:
That's great, didn't know about that :)

MrSerth (approved):
Nice, that's really helpful. Thanks for updating our documentation!

PR #232, opened 2022-10-13
Hide Nomad specific environment variables

### LINKED ISSUE #41: Hide system environment variables / Add toggle
With #27 and #39, we allowed access to (all) system environment variables with the intention to use those defined in the `Dockerfile`. Simultaneously, all other system variables are now accessible, especially those specific to our Nomad environment.

As outlined in [my comment](https://github.com/openHPI/poseidon/pull/39#issuecomment-971442550) I don't see an immediate call to action here but still room for improvement. Hence, we might investigate how to hide some system variables and could also think of allowing specific variables or introducing an HTTP query parameter such as `showSystemEnvironmentVariables` (see [proposal here](https://github.com/openHPI/poseidon/pull/39#issue-1055898994).

mpass99:
Closes #41 

... from the user environment.
Also refactor wrapping with "sh", "-c".

### DIFF
--- a/api/swagger.yaml
+++ b/api/swagger.yaml
@@ -497,7 +497,7 @@ paths:
               type: object
               properties:
                 command:
-                  description: The command to be executed. The working directory for this execution is the working directory of the image of the execution environment
+                  description: The command to be executed. The working directory for this execution is the working directory of the image of the execution environment. Single quotation ' can not be used.
                   type: string
                   example: python exercise.py
                 privilegedExecution:
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -105,7 +105,7 @@ In order to function properly, Poseidon expects the following commands to be ava
 - `ls`
 - `mkfifo`
 - `rm`
-- `sh` (not compatible with `zsh`)
+- `bash` (not compatible with `sh` or `zsh`)
 - `sleep`
 - `tar` (including the `--absolute-names` option)
 - `true`
--- a/internal/api/runners.go
+++ b/internal/api/runners.go
@@ -14,6 +14,7 @@ import (
 	"net/http"
 	"net/url"
 	"strconv"
+	"strings"
 )
 
 const (
@@ -30,6 +31,8 @@ const (
 	RecursiveKey            = "recursive"
 	PrivilegedExecutionKey  = "privilegedExecution"
 )
+
+var ErrForbiddenCharacter = errors.New("use of forbidden character")
 
 type RunnerController struct {
 	manager      runner.Accessor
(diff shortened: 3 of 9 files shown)


MrSerth:
I changed the status of this PR to "draft" as we agreed on a potential solution a week ago that needs to be tackled.

mpass99:
> * Switch to bash with single quotes

Two fun facts about that:
- Nested quotation does not work by escaping but with concatenating strings: `bash -c 'bash -c '"'"'echo Hi'"'"''`
- While single quotations work in our shell environments, it does not work directly with Nomad Exec. If we request `["bash", "-c", "'echo Hi'"]` we get `No such file or directory` for the `'echo Hi'` part. So we have to not enquote the outer bash payload but only the inner bash payload: `["bash", "-c", "/sbin/setuser user bash -c 'echo Hi'"]`.

MrSerth:
> Current state: The e2e tests are working in the local environment but not with Staging Nomad or in the CI.

That's already great progress and I am sure we'll find the reason for the failing tests 💪 

Do you think the integration of AWS could raise those failing tests? In your failing test, [one line reads](https://github.com/openHPI/poseidon/actions/runs/3353380422/jobs/5556123490#step:14:57): 
```
bash: unset "${!AWS@}" && make run: command not found
```

That makes sense for AWS: We don't have make available there and the environment variable hidden is also related to the AWS environment. Maybe our custom make command substitution is not working as expected?

I just pushed a [temporary change to the CI](https://github.com/openHPI/poseidon/commit/bc26a0395f906fdcacf16c1c68464ed66e5e7bee) and disabled AWS for a test run and noticed the following:

1. It didn't have the same error with the `unset` && `make` command, confirming my assumption that this error is actually related to AWS
2. I discovered a [concurrent write to websocket connection](https://github.com/openHPI/poseidon/actions/runs/3354843025/jobs/5558687636#step:14:176) but wondered whether it is our source code?
3. AWS is still [not completely disabled in the test](https://github.com/openHPI/poseidon/actions/runs/3354843025/jobs/5558687636#step:14:298) and the [`isAWSEnvironment` might be `true`](https://github.com/openhpi/poseidon/blob/939f36dac6b6a2aabbf8543805af51c972ba335c/tests/e2e/environments_test.go#L22) without AWS support. I'd propose to fix this as well.

I hope this helps to get somewhat closer to the root cause ⚙️

mpass99:
> Do you think the integration of AWS could raise those failing tests?

You are right - great hint! The root cause was that fun fact two is also valid for AWS but I just adjusted it for Nomad...

MrSerth (approved):
Ah, glad my tip helped out to identify the issue and thanks for considering AWS as well ;)

PR #233, opened 2022-10-13
Fix/#204 influx idle runner

### LINKED ISSUE #204: Poseidon requires restart to send updated data to Influx

mpass99:
Closes #204 

This fixes two bugs that both lead to wrong data in the influx data points.

### DIFF
--- a/internal/api/environments_test.go
+++ b/internal/api/environments_test.go
@@ -87,10 +87,10 @@ func (s *EnvironmentControllerTestSuite) TestList() {
 
 	s.Run("returns multiple environments", func() {
 		call.Run(func(args mock.Arguments) {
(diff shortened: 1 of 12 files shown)


MrSerth (approved):
You're pushing my capabilities of understanding Go code and contexts once again... I've tried my very best to get it and think I finally got it. Independent from that: Let's deploy and observe it :)

PR #235, opened 2022-10-14
 Add retry-mechanism for sample, mark-as-used and return of Nomad runners

### LINKED ISSUE #142: Automatic outage recovery
Earlier today, we faced some DNS issues in our internal network. During that time, DNS queries were not answered correctly. As a consequence, Nomad was unreachable.

At the same time, some requests from CodeOcean still reached Poseidon (probably due to some name caching or whatever) which in turn had issues to handle these correctly. Of course, the application generated (unrecoverable) errors and logged these (such as issues with updating the file system, etc.). These errors occurred as long as the underlying (external) outage was not resolved.

However, during that outage, Poseidon also returned runners to CodeOcean. As Nomad was unreachable, these were not replaced:
```
May 11 16:50:49 poseidon poseidon[137547]: time="2022-05-11T16:50:49.620686Z" level=error msg="Couldn't create new runner for claimed one" environmentID=29 error="error registering new runner job: couldn't register runner job: error registering Nomad job: Put \"https://<nomad server>:4646/v1/jobs?namespace=poseidon\": dial tcp: lookup <nomad server>: Temporary failure in name resolution" package=environment
```

This problem led to another follow-up issue: 
1. We got out of sync with Nomad, as runners were claimed but (probably?) not synched to Nomad
2. Poseidon returned all runners from "its" pool without replacing them. The pre-warmed pool size was reduced to 0. Without an administrator manually syncing the environment, the system was in a dead-lock position: An empty pool and many requests. 

Therefore, I wondered: Can we enhance this outage recovery to reduce manual interventions (e.g., some retry mechanism or schedule sync requests if we have many errors for everything)? Or is there another clever solution you can think of?

I would consider this ticket to be of some importance to tackle, as these issues might reoccur.

mpass99:
May close #142 

This issue should resolve most occurrences of Sentry issue [POSEIDON-K](https://sentry.io/organizations/openhpi/issues/2818111536/?referrer=github_integration) and Sentry issue [POSEIDON-1G](https://sentry.io/organizations/openhpi/issues/3348237065/?referrer=github_integration).

The last occurrences of those issues are two months ago. Are the DNS issues resolved?


MrSerth:
I just rebased this branch in order to test it with the latest changes on Staging.

> Are the DNS issues resolved?

Kind of. We made some changes that should reduce the occurrence of these issues.

MrSerth (approved):
Nice, I just gave it several tests and was able to confirm a "self-healing" effect.

My test setup:
- Newest Poseidon built from this branch
- Working CodeOcean / Nomad / ... setup

I then went ahead and disabled DNS name resolution by design with `sudo service systemd-resolved stop` **and waited a few moments**. Then, I could request a new runner (though the API or when no database entry for the execution environment and user was found in CodeOcean). As expected, a new ID was returned and cannot be marked as used. When requesting new IDs (or manually deleting the runner IDs in CodeOcean's database and clicking run), I was able to clear the pool of available IDs. 

After a while, I turned on the name resolution again with `sudo service systemd-resolved start` and waited for the retry mechanism to kick in (depending on how long Nomad was unreachable, this might take a while). Slowly, the pool got refilled and the system started working again as expected.

From this perspective, I have the feeling that it worked quite well. I am just unsure about the timing and number of retries. If I read the code correctly, we are retrying each operation max. 18 times with exponentially larger time. This probably leads to $2^{18}s = 3.03days$? 

Depending on the issue, this might be okay, but also quite long. Let's just hope we won't need this redo mechanism too often (or at least not for such a long time 😁)

For now, I would go with this solution and observe the behaviour. In case we find it not sufficient (with regard to #142) or because more data get's out of sync, we could still think of a full re-think mechanism (e.g., once a day or week, similar to a restarted Poseidon service?)

PR #239, opened 2022-10-24
Catch context canceled error

mpass99:
Related to https://github.com/openHPI/codeocean/pull/1389#discussion_r988942608

I would propose to additionally extend the `ContentLengthWriter` (a small bit) to monitor the expected content length and actual length?!


MrSerth (approved):
Awesome, works flawlessly and I also like the addition of the monitoring data. 😁

PR #240, opened 2022-10-25
Support protected directories

### LINKED ISSUE #236: Leverage file system ACLs for copied files
As a _user_
I want to _create files in an existing subdirectory_
in order to _compile Java classes in packages_. 

<hr>

As an administrator, I have the following requirements:

- [ ] Copied files should neither be writable nor deletable by non-privileged users (but privileged users)
- [ ] Folders (and the original workspace) created during the copy operation should allow the creation of new files by non-privileged users (even when the folder was created by a privileged user)

<hr>

We could think of either using the immutable flag (`chattr +i`, requires `LINUX_IMMUTABLE` for the container) or ACLs (`setfacl`, requires file system and kernel support). 

I would start by looking into ACLs if possible, but am also open to any other suggestion. 

_Based on / Part of #226_

mpass99:
by setting the sticky bit to all explicitly requested directories.

Closes #236 

This MR requires explicit requests for (protected) directories in the update file system route.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,9 @@ LOWER_REPOSITORY_OWNER = $(shell echo $(REPOSITORY_OWNER) | tr A-Z a-z)
 E2E_TEST_DOCKER_CONTAINER := co_execenv_java
 E2E_TEST_DOCKER_TAG := 17
 E2E_TEST_DOCKER_IMAGE = "$(LOWER_REPOSITORY_OWNER)/$(E2E_TEST_DOCKER_CONTAINER):$(E2E_TEST_DOCKER_TAG)"
+# The base image of the e2e test image. This is used to build the base image as well.
+E2E_TEST_BASE_CONTAINER := docker_exec_phusion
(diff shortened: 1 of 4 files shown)


MrSerth:
My todo: Also update the Docker image and ...
- [ ] set the sticky bit for `/workspace`
- [ ] reset the owner for `/workspace` to `root`

MrSerth:
I rebased the branch (to include the latest changes from `main`) and thereby also triggered a new build including the changed Dockerfile from https://github.com/openHPI/dockerfiles/pull/24. Unfortunately, I had to change the Makefile as well, so that the changes in the base image are applied. See #243.

MrSerth (commented):
Well done, I just tested the functionality and also added support for CodeOcean in https://github.com/openHPI/codeocean/pull/1410

With that change, I can now execute Java exercises with packages but not delete those files copied:

Bildschirm­foto 2022-10-26 um 10 11 10


There is just a small typo I noticed. Furthermore, we might think of adding a documentation to the swagger.yml to indicate this changed behaviour?

MrSerth (approved):
Awesome, I am happy that we found such a great (and simple) solution. Thank you! 👏

Issue #244, opened 2022-10-26
Issue collection Grafana dashboards


MrSerth:
Open ToDos:
- [x] label text
- [x] Color mapping
- [x] flux code reusage (at least the envMapping)
- [x] `|> filter(fn: (r) => (not exists r.stage) or contains(value: r["environment_id"], set: ${environment_ids:json}))`
- [x] `Request Body Size` implies it is a generic request. In fact, this panel just displays the Update file system requests.
- [x] Panel for #242

_Originally posted by @mpass99 in https://github.com/openHPI/poseidon/issues/200#issuecomment-1279841257_

PR #245, opened 2022-10-26
Add color mapping to Grafana dashboard.

mpass99:
Related to #244 

#250 should be merged first


deploy/grafana-dashboard/utils/color_mapping.py
```diff
@@ -19,3 +21,9 @@ def color_mapping(name, color):
 
 
 grey_all_mapping = color_mapping("all", "#4c4b5a")
+
+color_mapping_environments = {}
+deep_update_dict(color_mapping_environments, color_mapping("p10/java:8-antlr", "yellow"))
+deep_update_dict(color_mapping_environments, color_mapping("p28/r:4", "blue"))
+deep_update_dict(color_mapping_environments, color_mapping("p29/python:3.8", "orange"))
+deep_update_dict(color_mapping_environments, color_mapping("p31/java:17", "red"))
```
MrSerth, line 29: What if ... we would change the code so that there is less repetition / duplication?

My first idea was to use something like

```python3
deep_update_dict(
    color_mapping("p10/java:8-antlr", "yellow"),
    color_mapping("p28/r:4", "blue"),
    # ...
)
```

my second idea was even more reduced (too much?):

```python3
{
    "p10/java:8-antlr": "yellow",
    "p28/r:4": "blue",
    # ...
}
# and then map each key/value pair to the required `color_mapping`
```

However, as dicts in Python and using a map function is not natively supported, one could either switch to a normal array:

```python3
mapping = [
    {"p10/java:8-antlr": "yellow"},
    {"p28/r:4": "blue"}
    # ...
]

map(convert, mapping) # where convert is a function taking one argument 
```

or just stick to the first proposal which might be easier. For a variable number of arguments, have a look at args and kwargs in Python.

deploy/grafana-dashboard/utils/color_mapping.py
```diff
@@ -19,3 +21,9 @@ def color_mapping(name, color):
 
 
 grey_all_mapping = color_mapping("all", "#4c4b5a")
+
+color_mapping_environments = {}
```
MrSerth, line 25: Should we include color mappings for all environments? Depending on the graph, the others might be green or just have another color

Bildschirm­foto 2022-11-10 um 00 13 57

MrSerth, line 25: Do you see any chance to decide for colors dynamically, e.g. based on the name? Let's consider, we would have a fixed color palette with ~20 sorted values. Then, we could just sort the names and map one color to each name / ID. This would be stable as long as the name is not changed and would still be dynamic so that new environments don't need to get reflected in the code. 

Just asking, probably not achievable (because you're making the color assignment here in Python before you know how many and which environments we have...)

MrSerth (commented):
Looking forward to stable colors 🎨 , I just have some minor questions but generally speaking, this looks good 🧑‍🎨

deploy/grafana-dashboard/utils/color_mapping.py
mpass99: Yeah, sounds good! What do you think about this `reduce` solution?

deploy/grafana-dashboard/utils/color_mapping.py
mpass99: > you're making the color assignment here in Python before you know how many and which environments we have...

Jup, I'm struggling overcome that. What do you think about the following solution:
- In this issue we read the environments from a text file and map them to `a fixed color palette with ~20 sorted values`.
- In another issue we create an ansible playbook that
  - Fetches the current environments from Poseidon and writes the text file
  - Generates and deploys the current dashboard

deploy/grafana-dashboard/utils/color_mapping.py
MrSerth: Ah, yes. I really like that!

deploy/grafana-dashboard/utils/color_mapping.py
MrSerth: Sounds good, let's do that!

mpass99: See [3b168aa](https://github.com/openHPI/poseidon/pull/245/commits/3b168aaf905cf0f943fbd629124e4886a659c622) and #255

deploy/grafana-dashboard/utils/color_mapping.py
```diff
@@ -19,3 +23,13 @@ def color_mapping(name, color):
 
 
 grey_all_mapping = color_mapping("all", "#4c4b5a")
+color_mapping_environments = {}
+colours = ["yellow", "blue", "orange", "red", "purple",
+           "pink", "brown", "black", "white", "gray",
+           "gold", "super-light-red", "dark-red", "dark-orange", "super-light-yellow",
+           "super-light-green", "dark-green", "dark-blue", "super-light-purple", "super-light-blue"]
```
MrSerth, line 30: What do you think of using colors from the [R ColorBrewer](https://r-graph-gallery.com/38-rcolorbrewers-palettes.html)?

```suggestion
colours = [
            "#E41A1C", "#377EB8", "#4DAF4A", "#984EA3", "#FF7F00", "#FFFF33", "#A65628", "#F781BF", "#999999", # R ColorBrewer Set1
            "#8DD3C7", "#FFFFB3", "#BEBADA", "#FB8072", "#80B1D3", "#FDB462", "#B3DE69", "#FCCDE5", "#D9D9D9", "#BC80BD", "#CCEBC5", "#FFED6F" # R ColorBrewer Set3
]
```

deploy/grafana-dashboard/utils/color_mapping.py
```diff
@@ -19,3 +23,13 @@ def color_mapping(name, color):
...
+
+with open("environments.json") as f:
+    environments = json.load(f)
```
MrSerth, line 33: I would propose sorting the list, so that colors are more independent of the sorting in the JSON (and stable)

```suggestion
    environments = json.load(f)
    environments.sort()
```

deploy/grafana-dashboard/utils/color_mapping.py
MrSerth: Nice, I like it, thanks! I just have some color suggestions:

Current version:

Bildschirm­foto 2022-11-26 um 13 23 47

R ColorBrewer:

Bildschirm­foto 2022-11-26 um 13 27 38

Sorted R ColorBrewer:

Bildschirm­foto 2022-11-26 um 13 28 27


I deployed that to "Poseidon autogen Seb", so feel free to have a look :)

MrSerth (approved):
Finally, we have stable colors in Grafana. That's so awesome, looking forward to many great use cases for the dashboard!!

Issue #82, opened 2021-12-22
Add nomad installation and configuration to our ansible playbook


MrSerth:
In #79, we improved the documentation for configuring Nomad for Poseidon. Another way to document these requirements would be to add the installation of Nomad (and required plugins) to our ansible playbook. 

_Originally posted by @mpass99 in https://github.com/openHPI/poseidon/pull/79#discussion_r773921190_

mpass99:
The ansible playbook [here](https://github.com/openHPI/poseidon/issues/107#issuecomment-1072785494) may help implementing the automated Nomad configuration.

MrSerth:
In the past days, I spent a lot of time to finally move this ticket forward. After finishing the resource creation via Terraform, I now also tackled the provisioning using Ansible:

- By today, I am using a program called [terraform-inventory](https://github.com/adammck/terraform-inventory) to query the inventory for Ansible dynamically from the Terraform state. 
- Based on this inventory, I am running multiple playbooks to the different hosts and not only covered Nomad (with agents and servers), but also the generation of certificates (for the CA and the browser) and Poseidon. 
- Additionally, a GitLab CI/CD pipeline allows the execution of all steps remotely without a local setup.

<hr>

In order to close this issue, I would kindly ask you to:

- [ ] Checkout the repo locally:
  - [ ] Is the documentation clear?
  - [ ] Can you install the required dependencies and execute 
    - [ ] Terraform
    - [ ] Ansible
  - [ ] Did I integrate everything from our the playbooks you created previously? (I used them as a base for the ones I create)
- [ ] Test the GitLab CI pipeline:
  - [ ] Can you run it?
  - [ ] Is the pipeline clear?
- [ ] Test the deployment:
  - [ ] Is everything working as expected? I deployed (Nomad + Poseidon) to Staging with the `-terraform` suffix. You may choose to remove and redeploy there (of course, you might also adapt the count value or other variables)  :) 
  - [ ] Did I miss anything?

Please let me know what you think; looking forward to your feedback! Once you approved the changes (and we close this ticket), I'll switch to a Terraform + Ansible-managed system for Staging and Production.

mpass99:
It's awesome! Please have a look at `codeocean-terraform/53` for the review

MrSerth:
We finalised our playbooks and the Terraform integration and therefore can close this issue.

PR #248, opened 2022-11-07
Change to new ansible deployment process

mpass99:
Related to #82

> Did I integrate everything from our playbooks you created previously? (I used them as a base for the ones I create)

Yes. The new playbooks are great! I learned a lot about how structured playbooks can be :star_struck:


MrSerth (approved):
>> Did I integrate everything from our playbooks you created previously? (I used them as a base for the ones I create)
>
> Yes. The new playbooks are great! I learned a lot about how structured playbooks can be 🤩

Happy to hear! I am fine with switching, but would wait for the final review before merging here. 🙂

PR #250, opened 2022-11-09
Change Grafana panel label structure

mpass99:
by moving the stage and environment id to the front.

For the environment specific panels, the information about the stage, the environment id and the human readable image is important. I could not find another way to display those information apart concatenating them. I thought about setting the [display name](https://grafana.com/docs/grafana/latest/panels-visualizations/configure-standard-options/) to a specific column and assumed other labels/columns would be displayed by hovering, but I could not confirm that.

To do a small readability improvement, I moved the stage and environment id information to the front.

Related to #244


MrSerth (approved):
Good idea, with those small labels that's really easier to grab. It's somewhat sad that the display name is not working, but let's continue without worrying too much about that for now.

PR #251, opened 2022-11-09
Refactor environment mapping

mpass99:
Related to #244 
Merge conflict with #250 

This PR removes the 12 times duplication of the environment mapping.
The environment mapping maps the environment id to a label consisting of the stage, the environment id and the image.


MrSerth (dismissed):
Ah, nice! That really comes in handy, I was already thinking of something similar for #250. Perfect timing!

PR #252, opened 2022-11-09
Fix wrongly filtered data

mpass99:
by not assuming that every duration data point has a environment id tag.

Related to #244


MrSerth (approved):
Good catch, I probably wouldn't have noticed that without your PR!

PR #253, opened 2022-11-09
Feature/#244 panel files

mpass99:
Related to #244
Depends on #245 and #251

In a small try I was not able to download less that everything from Poseidon via CodeOcean even if I aborted my download. I achieved that only with Postman instead of CodeOcean. Maybe we need to observe that.


deploy/grafana-dashboard/queries/environment-mapping.flux
```diff
@@ -1,3 +1,4 @@
+
```
MrSerth, line 1: where is this line coming from? 😉

MrSerth (approved):
I like it, thanks! Also, I fixed the CodeOcean bug with https://github.com/openHPI/codeocean/commit/60f8d9809bd25ac1779f485a1c748c4a9a6dec9a. In development, the version I used worked, but in production we seem to have more race conditions, so I switched that.

deploy/grafana-dashboard/queries/environment-mapping.flux
mpass99: It bothered me that there was no free line before the environment mapping part and I didn't want to add a line at the bottom of all the other queries :sweat_smile:

MrSerth: Ah, I see! What if we modify the `read_query` method to add an empty line there? This would leave the flux files normally and still have the combined result look better.

MrSerth (approved):

PR #264, opened 2022-12-27
Document the container updater mechanism.

### LINKED ISSUE #208: Document container update mechanism
We should add some documentation on the update mechanism we have for execution environments. Especially interesting are the following questions:

- When are we destroying all unused runners?
- How do we handle an update to the resource constraints (CPU, memory)?
- What happens when adjusting the pre-warming pool size?
- Is there something specific for network access (enabling or disabling it, exposed ports)?

mpass99:
Closes #208

### DIFF
--- a/docs/nomad_usage.md
+++ b/docs/nomad_usage.md
@@ -34,3 +34,11 @@ When a user requests a runner, a runner from this pool can be used.
 In the background, a new runner is created, thus replenishing the pool.
 By running in the background, the user does not have to wait as long as the runner needs to start.
 The implementation of this concept can be seen in [the Runner Manager](/internal/runner/manager.go).
+
+### Lifecycle
+
+The prewarming pool is initiated when a new environment is requested/created according to the requested prewarming pool size.
+
+Every change on the environment (resource constraints, prewarming pool size, network access) leads to the destruction of the environment including all used and idle runners (the prewarming pool). After that, the environment and its prewarming pool is re-created.
+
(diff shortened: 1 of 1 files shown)


docs/nomad_usage.md
```diff
@@ -34,3 +34,11 @@ When a user requests a runner, a runner from this pool can be used.
...
 By running in the background, the user does not have to wait as long as the runner needs to start.
 The implementation of this concept can be seen in [the Runner Manager](/internal/runner/manager.go).
+
+### Lifecycle
+
+The prewarming pool is initiated when a new environment is requested/created according to the requested prewarming pool size.
+
+Every change on the environment (resource constraints, prewarming pool size, network access) leads to the destruction of the environment including all used and idle runners (the prewarming pool). After that, the environment and its prewarming pool is re-created.
+
+Other causes of the destruction of the prewarming pool are the explicit deletion of the environment by using the API route and via the get environment route with the fetch flag and Nomad not having the corresponding template job.
```
MrSerth, line 44: I find it difficult to get the second part of this sentence. Sure, I am aware of the `fetch` flag, which will query Nomad for the actual prewarming pool size instead of using the cached stats from Poseidon. However, I don't get the scenario in which this relates to a change of the Docker images (i.e., downloading the image to create a template job). Besides that, it reads as this would destruct the prewarming pool (both parts of the sentence relate to "Other causes of the destruction...", don't they?).

MrSerth (approved):

PR #266, opened 2022-12-28
Fix/#249 1 q list file system

mpass99:
See #249

This may be a race condition between the runner removal and the `listFileSystem` route. To reduce the time span of this condition we first remove the runner from the internal `usedRunner` representation as then the `listFileSystem` is more likely to return `RunnerNotFound`.
If this error appears again, we got more debug information (the runner id) to take a closer look.


MrSerth (approved):
Sure, let's give this a try. I'll trigger a rollout in a bit to include the logging 🤞

MrSerth:
I deploed Poseidon to production.

PR #275, opened 2023-02-02
Sentry Performance Tracing

### LINKED ISSUE #270: Investigate performance of Score Runs
Some score runs triggered in CodeOcean are notoriously slow and take quite some time to finish. This led to some teachers spending a lot of time combining various test cases to minimize scoring results. One of the optimizations taken is to squash all test cases into one file, don't use the linter (in Python) or don't use a testing framework at all. Especially putting all tests into one file is somewhat understandable, as in the current workflow test files get executed sequentially.

Obviously, we want to optimize our tool rather than putting more efforts on teachers to optimize their tests. Therefore, before taking any specific steps, we should get a better understanding of the score runs and actually collect some profiling data. The data could help us answer the following questions:

How much time is needed ...
- in CodeOcean
  - to save an exercise before the actual scoring request is issued
  - to collect files
  - to request a runner
  - to send files to Poseidon
- in Poseidon
  - to copy files to an allocation in Nomad
  - to prepare an execution
- in the allocation
  - to compile the code (if applicable, e.g., for Java)
  - to execute the actual test cases
- in CodeOcean
  - to parse the scoring result using the RegEx
  - store the output of the testrun (and other post-processing tasks)
  - to forward the response

Based on these numbers, we could identify the main pain points to tackle them individually.

mpass99:
Relates to #270 

With [Sentry's Distributed Tracing](https://docs.sentry.io/product/sentry-basics/tracing/distributed-tracing/) we can connect the Tracing of our applications to get a more detailed view into the time behavior.

For Poseidon we just have to enable the tracing functionality of Sentry. For that see also: [38ffc3e9](https://lab.xikolo.de/codeocean/codeocean-terraform/-/commit/38ffc3e9247abddf55191bf951569f6e79cd433d).

The tracing connection between the application is done automatically by Sentry (via the `trace-id` http header; [See](https://docs.sentry.io/platforms/ruby/performance/connect-services/)).

With only this we would still have two separate transactions - one for Submission create and one for Submission Score. To join them we can enable the Sentry Tracing for the JavaScript Frontend.


MrSerth:
Awesome, thanks for getting that started! I was reviewing commit-by-commit and find all changes plausible. Besides reviewing, I also started having a look at Sentry and was happy to see some traces there ;).

What I was still thinking are two things regarding the spans:

1. Should we add additional spans, e.g., in some filters running with every request or other often-used methods?
   This would allow us to see more details, such as shown in [this trace](https://sentry.io/organizations/openhpi/performance/codeocean:57c4621bae8c410090330525a21596c1).
2. Can we add the current (custom) span names into the "description" field and rather an "identifier" as the current name? This would allow to see the name for spans. For example, compare [this custom span](https://sentry.io/organizations/openhpi/performance/summary/spans/List%20File%20System:d41d8cd98f00b204/?project=6077585&transaction=GET+%2Fapi%2Fv1%2Frunners%2F%2A%2Ffiles) to [this pre-defined span](https://sentry.io/organizations/openhpi/performance/summary/spans/http.server:ea72d7f241df4722/?project=6077585&query=http.method%3APATCH&statsPeriod=24h&transaction=PATCH+%2Fapi%2Fv1%2Frunners%2F%2A%2Ffiles). We might also come up with an own naming scheme, e.g., `poseidon.operation`, `poseidon.manager`, `poseidon.storage`, `poseidon.nomad.execute`, `influx.send`, `nomad.query`, `response.render`, ...

mpass99:
Good remarks!  
I added an identifier for the custom spans.  
Also, I added more custom spans for the most used and most time consuming functionality, `Execute`. If you would like to know the time behavior of other methods, please let me know.

MrSerth (approved):
Awesome! I was just testing the new tracing with CodeOcean, and it works like a charm!

Let's evaluate the responses once we have some data to identify where to add more details. :)

PR #301, opened 2023-02-28
Add automatic detection for release version

MrSerth:
During the latest deployment, I noticed that we might run into issues with identifying the release information. Therefore, I changed the process to automatically grab the git hash used to build the binary. We can still specify a manual release version, e.g., to be used during development.


mpass99 (approved):
I remember when I saw this first reading the Go 1.18 Release Notes. Cool to have it in Poseidon :star_struck: 

We could think about adding behavior for the ` vcs.modified=true` setting :man_shrugging:

MrSerth:
> We could think about adding behavior for the  vcs.modified=true setting 🤷‍♂️

I also thought about this. Potentially, we could just append "-modified", as we've done before. However, as we currently build during our GitLab CI pipeline, we rarely have those "modified" builds.

mpass99:
I agree - not really important. But if we spot the prepended "modified", we might discover a new error..

MrSerth:
> I agree - not really important. But if we spot the prepended "modified", we might discover a new error..

Indeed. I changed the code accordingly.

mpass99 (approved):

PR #302, opened 2023-02-28
Actual retry sending InfluxDB data

### LINKED ISSUE #262: Could not write influx data.
We thought of wrapping the transaction and retrying it (at least once) if it failed. Then, we should only log an error if sending data failed repeatedly.

---

Sentry Issue: [POSEIDON-1H](https://sentry.io/organizations/openhpi/issues/3399850901/?referrer=github_integration)

```
Could not write influx data.
```

MrSerth:
Previously, we always logged the error on first failure and (nevertheless) tried to send the data within 3 minutes (default configuration).

Fixes POSEIDON-1H
Closes #262


mpass99 (approved):
Great, thank you

PR #321, opened 2023-03-15
Fix multiple user Runner use

mpass99:
Related to #269

When deploying (Poseidon/) Nomad, we can [observe in the logs](https://github.com/openHPI/poseidon/files/10978226/syslog.suspect.poseidon.txt) that first some Nomad-specific failures occur and after that all runner get added again.

When adding the runner (again), there was no check if the runner is already in use. Therefore, the already used runner is re-added to the idleRunner. If this (/those) runner is not reassigned to another user before the runner is destroyed (e.g. due to its timeout) there is no problem. But, if the runner (A) gets re-provided to another user, another runner (B) gets started, but in the end the runner (A) will be removed just once. When trying to remove the runner (A) the second time only an error (`DeleteJob failed: job not found`) is created. In total we have now (one) more runner and our isolation constraint failed.

For this moment it stays unexplained, why Nomad is re-adding the runner because on a normal systemd restart, Poseidon throws the error `Stopped updating the runners! Retry 1" error="error receiving events: unexpected EOF` and does not re-add all the runner. I assume the difference is that in the deployment the nomad-server stays active and just the nomad-agents (or even just their docker service) are reloading.

I've added a ~~Sentry~~ Grafana Alert to ensure that we will know when this issue happens again.

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -9,7 +9,9 @@ import (
 	"github.com/openHPI/poseidon/internal/config"
 	"github.com/openHPI/poseidon/pkg/dto"
 	"github.com/openHPI/poseidon/pkg/logging"
+	"github.com/openHPI/poseidon/pkg/monitoring"
 	"github.com/openHPI/poseidon/pkg/nullio"
+	"github.com/openHPI/poseidon/pkg/storage"
 	"io"
 	"strconv"
 	"strings"
@@ -35,6 +37,12 @@ type AllocationProcessoring struct {
 }
 type AllocationProcessor func(*nomadApi.Allocation)
 type AllocationProcessorMonitored func(*nomadApi.Allocation, time.Duration)
+
+type allocationData struct {
+	// allocClientStatus defines the state defined by Nomad.
+	allocClientStatus string
+	start             time.Time
+}
 
 // ExecutorAPI provides access to a container orchestration solution.
 type ExecutorAPI interface {
@@ -184,14 +192,15 @@ func (a *APIClient) WatchEventStream(ctx context.Context, callbacks *AllocationP
 	if err != nil {
 		return fmt.Errorf("failed retrieving allocation stream: %w", err)
 	}
-	pendingAllocations := make(map[string]time.Time)
+	// allocations contain management data for all pending and running allocations.
+	allocations := storage.NewMonitoredLocalStorage[*allocationData](monitoring.MeasurementNomadAllocations, nil, 0, nil)
 
 	handler := func(event *nomadApi.Event) (bool, error) {
 		switch event.Topic {
 		case nomadApi.TopicEvaluation:
 			return false, handleEvaluationEvent(a.evaluations, event)
 		case nomadApi.TopicAllocation:
-			return false, handleAllocationEvent(startTime, pendingAllocations, event, callbacks)
+			return false, handleAllocationEvent(startTime, allocations, event, callbacks)
 		default:
 			return false, nil
 		}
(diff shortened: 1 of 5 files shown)


mpass99:
We would introduce a race condition: A ProvideRunner request could claim the runner between line 149 and 166. 

To fix this we might need to add a Lock functionality to the storage interface?!

MrSerth:
Wow, thanks for looking into this and finding such a good explanation for the issue. Yes, we probably should fix that to ensure one runner only gets assigned to one user. 

If I get your changes right, you are preventing to assign an already used allocation to the pool of idle runners (by checking in the `if` statement) and also add a Sentry event in this case? This means, we won't see the "failed to scale down error" but rather the new one with the same frequency instead? 

I also see the issue of a race condition and sure, we could add a lock (that would make sense, I think). Other than that, I also wondered whether we should (temporarily) delay returning new runner IDs while we process that many events from the event stream? I mean, if we noticed that Nomad hat such a severe error and (almost all) allocations are re-added, cannot we wait until all events have been processed before claiming new runners? Maybe, this is just another way to solve the race condition, but I was wondering how we could minimize the synchronization efforts. Otherwise, just proceed with your idea; I would argue that ensuring to return a runner only once is the upmost goal.

mpass99:
> If I get your changes right, you are preventing to assign an already used allocation to the pool of idle runners (by checking in the `if` statement)

Yes

>  and also add a Sentry event in this case?

No, only a debug statement by now. (Let's maybe elevate this to a Sentry error depending on the following discussion)

> This means, we won't see the "failed to scale down error" but rather the new one with the same frequency instead?

The difference is that the `failed to scale down` error appears on Poseidon restarts, while this error would appear when Nomad re-adds the runner. But, maybe we can avoid both errors:

> I also see the issue of a race condition and sure.

Both solutions appeared to be not so trivial when it comes to the implementation. Also, they only handle the symptoms. Therefore, I took another look on the cause that make Nomad sending the events. I was able to reproduce the behavior by restarting the Docker service. And it makes sense - When restarting Docker all running container get removed. After the Docker restart, Nomad restarts the containers and sends us the events of the started Allocations (/ Tasks). When taking an attentive look at the events Nomad sends us, we notice that the counter `alloc.TaskStates["default-task"].Restarts` is increased. This leads to a third solution, we can check the number of restarts of an allocation. Then we handle the allocation removal when the container is stopped and the allocation placement when the container is restarted. An implementation for this is sketched in [a3ac64b](https://github.com/openHPI/poseidon/pull/321/commits/a3ac64b6cb2281e386f7f6221afa4560c981b178).

MrSerth:
As discussed, I am fine with sending that error to Sentry. We should just evaluate which information we need in order to debug it afterwards (and attach that information before sending the event).

Further, we discussed the restart counter and the use. Potentially, it can be removed, since an allocation int the runningAllocation map has been started already ;). When changing these lines, you also indicated the wish to refactor the code to use the storage backend :+1:.

mpass99:
This PR is pending since we found indicators that the error might have other causes.

internal/nomad/nomad.go
```diff
@@ -256,8 +265,8 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 // If a new allocation is received, onNewAllocation is called. If an allocation is deleted, onDeletedAllocation
 // is called. The pendingAllocations map is used to store allocations that are pending but not started yet. Using the
```
MrSerth, line 266: This comment seems to be outdated, as there is no `pendingAllocations` map any longer.

internal/nomad/nomad_test.go
```diff
@@ -534,7 +536,10 @@ func TestAPIClient_WatchAllocationsReturnsErrorWhenAllocationCannotBeRetrievedWi
 		Type:  structs.TypeAllocationUpdated,
 		Topic: nomadApi.TopicAllocation,
 		// This should fail decoding, as Allocation.ID is expected to be a string, not int
-		Payload: map[string]interface{}{"Allocation": map[string]interface{}{"ID": 1}},
+		Payload: map[string]interface{}{"Allocation": map[string]interface{}{
+			"ID":         1,
+			"TaskStates": map[string]interface{}{TaskName: map[string]interface{}{"Restarts": 0}},
```
MrSerth, line 541: Is this information on the restarts here and in line 639 still needed, or a left-over?

MrSerth (commented):
Nice, thank you! If I read this correctly, you also switched to use the generic Storage class, right? 👏 

I just have two minor questions, but otherwise think it's great! 🦚

mpass99:
Thank you!

After the deployment, we should also synchronize all environments.

MrSerth (approved):

MrSerth (approved):
I just rebased your changes, will deploy them and synchronize all environments afterward.

MrSerth:
I deployed and synced all runners. Once observation I noticed is that [POSEIDON-2D](https://codeocean.sentry.io/issues/4052093496/?project=6077585&query=is%3Aunresolved&referrer=issue-stream&stream_index=0) occurred: "Removing not listed allocation". Is that expected and okay?

Issue #328, opened 2023-03-20
Oscillating Idle Runner Pool


MrSerth:
In our monitoring, we recently discovered that the idle runner pool size is oscillating between zero and another, static number of runners. This happens even though the given execution environment has not been used recently. We would like to learn why we observe this effect and would be interested to fix it.

 
<img width="894" alt="Bildschirm­foto 2023-03-20 um 17 42 25" src="https://user-images.githubusercontent.com/7300329/226409321-87934f0c-b36a-47f1-be49-5d5f07974960.png">

PR #330, opened 2023-03-21
Fix Panic Recovery

mpass99:
Related to #328 

Our Sentry panic handler was never called as the panic is thrown in a separate goroutine [[1]](https://stackoverflow.com/questions/46230973/how-to-recover-from-an-asynchronous-panic-in-external-package) [[2]](https://go.dev/blog/defer-panic-and-recover).
By moving the recovery functionality in the main goroutine, now panics are captured as Sentry events.

### DIFF
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -110,7 +110,10 @@ func initProfiling(options config.Profiling) (cancel func()) {
 	return cancel
 }
 
-func runServer(server *http.Server) {
+func runServer(server *http.Server, cancel context.CancelFunc) {
+	defer cancel()
+	defer shutdownSentry() // shutdownSentry must be executed in the main goroutine.
+
 	log.WithField("address", server.Addr).Info("Starting server")
 	var err error
 	if config.Config.Server.TLS.Active {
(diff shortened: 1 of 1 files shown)


cmd/poseidon/main.go
```diff
@@ -213,15 +221,15 @@ func main() {
 	}
 	logging.InitializeLogging(config.Config.Logger.Level)
 	initSentry(&config.Config.Sentry, config.Config.Profiling.Enabled)
-	defer shutdownSentry()
 
 	cancel := monitoring.InitializeInfluxDB(&config.Config.InfluxDB)
```
MrSerth, line 225: Should we rename this occurrence of `cancel` (and in the line below) to avoid confusion with the new one introduced in line 231?

cmd/poseidon/main.go
```diff
@@ -193,11 +196,16 @@ func initServer() *http.Server {
...
 // shutdownOnOSSignal listens for a signal from the operating system
 // When receiving a signal the server shuts down but waits up to 15 seconds to close remaining connections.
-func shutdownOnOSSignal(server *http.Server) {
+func shutdownOnOSSignal(server *http.Server, ctx context.Context) {
 	// wait for SIGINT
 	signals := make(chan os.Signal, 1)
 	signal.Notify(signals, syscall.SIGINT, syscall.SIGTERM)
-	<-signals
+	select {
+	case <-ctx.Done():
```
MrSerth, line 204: Just for my understanding: the context will be `Done` when cancelled, i.e., when a panic occurs? Otherwise, Poseidon will run until a signal is received?

And shouldn't we include lines 209-215 before closing the select in line 207 (so that it's clearer what happens there)?

mpass99: > the context will be `Done` when cancelled, i.e., when a panic occurs?

Yes, when our http server stops (due to a panic or an error).

> Otherwise, Poseidon will run until a signal is received?

Yes.

> And shouldn't we include lines 209-215 before closing the select in line 207 (so that it's clearer what happens there)?

Yeah ok :+1:

MrSerth (approved):

PR #331, opened 2023-03-21
Fix Concurrent Map Write

mpass99:
Related to #328 

We created a race condition when handling the Sentry Debug Messages asynchronously. Two debug messages can be evaluated at the same time leading to the following log:
``` log
fatal error: concurrent map writes
goroutine 4361 [running]:
github.com/getsentry/sentry-go.(*Span).SetData(...)
#011/home/poseidon/go/pkg/mod/github.com/getsentry/sentry-go@v0.19.0/tracing.go:236
github.com/openHPI/poseidon/internal/nomad.(*SentryDebugWriter).handleTimeDebugMessage(0xc00022e420, 0x0?)
#011/opt/poseidon/internal/nomad/sentry_debug_writer.go:98 +0x285
created by github.com/openHPI/poseidon/internal/nomad.(*SentryDebugWriter).Write
#011/opt/poseidon/internal/nomad/sentry_debug_writer.go:67 +0x19b
goroutine 1 [chan receive, 2 minutes]:
main.shutdownOnOSSignal(0x18f3900?)
#011/opt/poseidon/cmd/poseidon/main.go:200 +0x97
main.main()
#011/opt/poseidon/cmd/poseidon/main.go:226 +0x15a
```

In this context we noticed that the many panics are not captured by Sentry. See #330

The oscillation is due to Poseidon is restarting, first recovering the environments (with 0 idle runner) & monitors the number of idle runner, and then loading the (idle) runner.

### DIFF
--- a/internal/nomad/sentry_debug_writer.go
+++ b/internal/nomad/sentry_debug_writer.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"regexp"
 	"strconv"
+	"sync"
 	"time"
 )
 
@@ -27,9 +28,10 @@ var (
 // For matches, it creates a Sentry Span. Otherwise, the data will be forwarded to the Target.
 // The passed context Ctx should contain the Sentry data.
 type SentryDebugWriter struct {
-	Target   io.Writer
-	Ctx      context.Context
-	lastSpan *sentry.Span
+	Target       io.Writer
+	Ctx          context.Context
+	lastSpan     *sentry.Span
+	lastSpanLock sync.Mutex
 }
 
 func NewSentryDebugWriter(target io.Writer, ctx context.Context) *SentryDebugWriter {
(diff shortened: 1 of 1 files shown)


mpass99:
It is deployed and we have no oscillation anymore. Anyways, I would appreciate your review. Maybe it would be better to just not do it asynchronously.. Can we be sure about the order of the sentry spans this way (race condition)..

MrSerth (commented):
Thanks for analyzing and fixing this bug together with deploying these changes! :+1:

I think your approach is fine, and locking might be the only way to handle it here (besides making everything synchronous, which might have a negative impact on the overall performance). As long as we ensure that we are not mixing two spans (e.g., one span got started and then updated here with the timing, or description of another span), I am completely fine. 

According to the Sentry docs, the order is not important, as Sentry will re-order them based on the time stamps. Therefore, I wouldn't worry about the actual order, as long as the information is correct.

MrSerth:
Ah, I see. Thanks for explaining your thoughts in detail.

I haven't seen that as well, but agree that it could occur at any time. The only solution I can currently think of is not to hold just the reference to the _last_ span. Theoretically, we could hold a reference to the array of all spans, add an (empty) span when preparing the command with the debug statement and include the array index of this new span to the message, and use this one to later identify the span and just add the timing / description.

PR #336, opened 2023-03-24
Add context to Sentry Hook

### LINKED ISSUE #326: Associate Sentry Errors with Sentry Transactions
Today, we noticed that we don't see the Sentry transaction when viewing Sentry errors. Making a link between both would, potentially, make it easier to debug issues.

mpass99:
Closes #326 

With updating our Sentry Hook, tracing information stored in the passed context can be associated with sentry events/issues.

With this first commit, we added the context just to log statements where an context was available. In many places we do not pass the (request) context. I think it would be a good practice to do so in the future, (not only) because we then could associate Sentry issues with Sentry Traces. We may use this PR for a refactoring into the direction of more context passing.

I noticed, that Sentry developed [its own Logrus Hook](https://github.com/getsentry/sentry-go/blob/master/logrus/logrusentry.go) 4 months ago. We may think about switching to that. At this moment their implementation does not support the tracing association but there is an open PR https://github.com/getsentry/sentry-go/pull/522

### DIFF
--- a/internal/api/api.go
+++ b/internal/api/api.go
@@ -42,7 +42,7 @@ func NewRouter(runnerManager runner.Manager, environmentManager environment.Mana
 func configureV1Router(router *mux.Router,
 	runnerManager runner.Manager, environmentManager environment.ManagerHandler) {
 	router.NotFoundHandler = http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
-		log.WithField("request", r).Debug("Not Found Handler")
+		log.WithContext(r.Context()).WithField("request", r).Debug("Not Found Handler")
 		w.WriteHeader(http.StatusNotFound)
 	})
 	v1 := router.PathPrefix(BasePath).Subrouter()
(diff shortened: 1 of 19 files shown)


MrSerth (commented):
Nice, thanks for continuing your efforts to get Sentry right.

I am still wondering about the need to pass the request context just everywhere (shouldn't that just be available "globally"?), but I am fine doing so. One idea to reduce the need to pass the context as a parameter could be to attach it as an attribute to the objects that we use...

I was just checking the list of log occurrences and thought some additional might get a context. Precisely, I thought of the following: 

- [x] `internal/api/api.go:45`
- [x] `internal/api/websocket.go:50`
- [x] `internal/api/websocket.go:65` # might be more difficult
- [x] `internal/api/websocket.go:71` # might be more difficult
- [x] `internal/api/websocket.go:100`
- [x] `internal/api/ws/codeocean_reader.go:100`
- [x] `internal/api/ws/codeocean_reader.go:113`
- [x] `internal/api/ws/codeocean_writer.go:130`
- [ ] `internal/api/ws/codeocean_writer.go:142`
- [ ] `internal/api/ws/codeocean_writer.go:146`
- [ ] `internal/api/ws/codeocean_writer.go:150`
- [ ] `internal/nomad/api_querier.go:94`
- [ ] `internal/nomad/api_querier.go:132`
- [ ] `internal/nomad/api_querier.go:135`
- [ ] `internal/runner/nomad_runner.go:291`
- [ ] `internal/runner/nomad_runner.go:293`

Sure, not every occurrence of `log` is a good indicator, as some tasks happen asynchronously outside a request (such as then handling of Nomad events when a new runner was started). I tried filtering the list above; and some might not be achievable easily (which you might want to skip), but others might work quite easily.

Finally, I was wondering whether we need to extract the transaction ID ourselves. In the Logrus integration you mentioned, their library is doing so (for example, see [here](https://github.com/getsentry/sentry-go/blob/231d5f0f44e706229b8f06c8257b1b603a52b691/logrus/logrusentry.go#L169-L178)). I might work already the current way; I haven't tested it. Once the PR you referenced is merged, we might also want to evaluate switching to their Logrus integration, as this will potentially be maintained by them in the future.

PR #345, opened 2023-04-02
Write performance profile on SIGUSR1

MrSerth:
Building Poseidon with a performance profile (as introduced in #297) might take quite long, actually more than five minutes. As we previously discovered that the profile needs to be written completely (what a surprise), we decided to stop Poseidon before starting to build the new version. This, of course, flushed the performance profile, but also rendered our service unusable during the build time.

With these changes, we allow Poseidon to write the performance profile on the `USR1` signal. This can be sent as part of the deployment process, thus reducing the downtime of Poseidon.


cmd/poseidon/main.go
```diff
@@ -196,16 +196,28 @@ func initServer() *http.Server {
 
 // shutdownOnOSSignal listens for a signal from the operating system
 // When receiving a signal the server shuts down but waits up to 15 seconds to close remaining connections.
-func shutdownOnOSSignal(server *http.Server, ctx context.Context) {
+func shutdownOnOSSignal(server *http.Server, ctx context.Context, stopProfiling func()) {
```
mpass99, line 199: What do you think about adding a test like
``` Go
func TestShutdownOnOSSignal_Profiling(t *testing.T) {
	called := false

	server := initServer()
	go shutdownOnOSSignal(server, context.Background(), func() {
		called = true
	})

	<-time.After(tests.ShortTimeout)
	err := syscall.Kill(syscall.Getpid(), syscall.SIGUSR1)
	require.NoError(t, err)
	<-time.After(tests.ShortTimeout)

	assert.True(t, called)
}
```

mpass99 (approved):
Good finding!

PR #356, opened 2023-04-11
Synchronize Sentry debug message handling.

### LINKED ISSUE #354: Refactor lastSpan for Sentry Debug Messages
In #331, we were discussing whether the order of debug messages is always guaranteed. Conclusion:

> I haven't seen that as well, but agree that it could occur at any time. The only solution I can currently think of is not to hold just the reference to the _last_ span. Theoretically, we could hold a reference to the array of all spans, add an (empty) span when preparing the command with the debug statement and include the array index of this new span to the message, and use this one to later identify the span and just add the timing / description.

_Originally posted by @MrSerth in https://github.com/openHPI/poseidon/issues/331#issuecomment-1493054276_

We should further observe and potentially refactor access to the last span, in order to avoid any issues.

mpass99:
Closes #354

In order to guarantee the order of the debug messages, we can handle the debug messages synchronous to the Write calls, as we expect the write calls to be in the right order.

We may even remove the Mutex?!

### DIFF
--- a/internal/nomad/sentry_debug_writer.go
+++ b/internal/nomad/sentry_debug_writer.go
@@ -7,7 +7,6 @@ import (
 	"io"
 	"regexp"
 	"strconv"
-	"sync"
 	"time"
 )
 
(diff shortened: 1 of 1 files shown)


MrSerth (commented):
> We may even remove the Mutex?!

I would argue we can do so. Since everything is now done synchronously, this shouldn't be a major problem. Otherwise, we re-add it? ;)

MrSerth (approved):

Issue #358, opened 2023-04-23
Unreliable prewarming pool size


mpass99:
The number of idle runner is varying strongly out of unknown reasons.

<details>
  <summary>The number of runner decreases.</summary>
  <p>

- On the 18th from 7am till 1pm decreases from 5 to 2 for environment 29 and from 5 to 4 for environment 10.
- On the 20th from 6:27 am to 4 pm environment 29 decreases from 11 to 3 runners
  - Sentry `Stderr task finished with error`
    - `error executing command in allocation: task "default-task" is not running.`
- On the 21st from 5am to 11am from 6 to 4
- On the 22nd from 3am to 9pm from 5 to 2
- On the 23rd environment 10 from 6am to 9 pm from 5 to 4

  </p>
</details>

<details>
  <summary>The number of runner increases.</summary>
  <p>

- On the 18th at 8pm environment 10 has one additional runner.
  - `Apr 18 19:55:18 Main process exited, code=exited, status=2/INVALIDARGUMENT`
  - `Apr 18 19:55:20 Started Poseidon`
  - Sentry `Couldn't scale environment`
- On the 19th at 6:30am an rapid increase to 8 idle runner (both environments).
- At 6:42am environment 29 goes to 12 runner
- At 6:46am environment 29 goes to 15 runner
- At 8:28am environment 29 increased by 1
  - `Apr 19 08:28:00 Main process exited, code=exited, status=2/INVALIDARGUMENT`
  - `Apr 19 08:28:02 Started Poseidon`
  - Sentry `Couldn't scale environment`
- On the 20th at 3am environment 10 from 4 to 7
  - `Apr 20 03:00:00 Stopping Poseidon`
  - `Apr 20 03:00:47 Started Poseidon`
  - Sentry `Couldn't scale environment`
  - 6x Sentry `Other Desired Status`
    - This warning is assumed to be related to a decreased runner count
    - Only template jobs and runner of environment 11 affected
    - Not related to this issue. See #357 instead
  - InfluxDB: Many allocations are added 7 or 8 times in the same second `03:01:50`
  - Nomad Server upgrade
- On the 21st at 3am environment 29 increased to 6
  - Sentry `Stderr task finished with error`
    - Nomad error `Unexpected response code: 500 (rpc error: failed to get conn: rpc error: lead thread didn't get connection)`
  - Sentry `Error executing the request`
[...]

Issue #357, opened 2023-04-17
Handle lost, pending allocations


mpass99:
In a recent downtime of one Nomad Agent we observed a new process we do not handle by now.

When runners are provided to users, another runner is started to refill the prewarming pool. In the process of being started, the allocation needs to be scheduled and started. In this time the allocation has the `ClientStatus` `pending`. When in this moment the Nomad agent  - that just starts the allocation -is going down, Nomad informs us with an event that the `ClientStatus` is `pending` but the `DesiredStatus` is `stop`. We convert the missing handling of this edge case into the warning `Other Desired Status` ([POSEIDON-2E](https://codeocean.sentry.io/issues/4052530553/?referrer=github_integration)).

**Open Question**: Why do we not receive an Event with the `ClientStatus` `lost` and the `DesiredStatus` `stop` like usual?

The downtime leads also to another error: `Started Runner is already in use` ([POSEIDON-2G](https://codeocean.sentry.io/issues/4061207219/?referrer=github_integration)). This can not be explained by the events that we have seen in the other warning as `pending` runners are not "started".

It shows us that also the runners with the `ClientStatus` `running` are lost but our warning about the `Other Desired Status` does not monitor this edge case.

Acceptance Criteria:
- [ ] Check for the missing Nomad events
- [ ] Handle lost, pending allocations
- [ ] Handle lost, running allocations and improve error detection.

MrSerth:
[This page](https://developer.hashicorp.com/nomad/docs/commands/node/status#examples) lists the relevant options for the Desired Status and Client Status. It also shows, as you suggested, that there might be a client status`lost`, which we haven't seen in our logs. I also don't know why we don't see it, but I noticed that the `pending` status is only sent once the Nomad agent is available again (not when the connection is lost).

mpass99:
> In a recent downtime of one Nomad Agent

In which? `Apr 17, 2023 8:56:49 AM UTC`

> Why do we not receive an Event with the ClientStatus lost and the DesiredStatus stop like usual?

I tested the scenario of an agent downtime on staging that was starting and running some allocations.
We can observe that for each starting and running allocations of that agent we receive a `lost` event: `Other Client Status` [POSEIDON-2M](https://codeocean.sentry.io/issues/4206026379/).

[...]

PR #359, opened 2023-04-23
Add Job ID to Nomad Allocation monitoring.

mpass99:
Related to #358 and #357 

In order to combine the monitoring information with the Poseidon logs we need to identify the allocation transactions with its job ID (and not only its allocation ID).


MrSerth (approved):
Sounds reasonable, thanks for digging into these errors!

PR #360, opened 2023-04-23
Add Client Status to Nomad Allocation monitoring.

mpass99:
It would also be great to see the client status in the monitoring data. Sorry for putting this in a separate PR :sweat_smile:

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -45,6 +45,8 @@ type allocationData struct {
 	allocClientStatus string
 	jobID             string
 	start             time.Time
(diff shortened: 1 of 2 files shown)


MrSerth:
Sure, don't worry. What about the desired status, or the node, ...?

mpass99:
> What about the desired status, or the node, ...?

The desired status is known as we check it specifically before adding or removing the allocation from the storage.
The node is a good catch and may simplify checking the Nomad logs if required :+1:

MrSerth (dismissed):

PR #372, opened 2023-05-11
Fix missing idle runners.

mpass99:
Related to #358 

The idle runner decrease from time to time as the Nomad event `AllocationUpdated` with the status `pending` is common but not always send and our process relied on it.

Automatic merge and deployment this night.

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -318,9 +318,6 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 // storage the state is persisted between multiple calls of this function.
 func handleAllocationEvent(startTime int64, allocations storage.Storage[*allocationData],
 	event *nomadApi.Event, callbacks *AllocationProcessoring) error {
-	if event.Type != structs.TypeAllocationUpdated {
-		return nil
-	}
 	alloc, err := event.Allocation()
 	if err != nil {
 		return fmt.Errorf("failed to retrieve allocation from event: %w", err)
@@ -359,12 +356,13 @@ func handleAllocationEvent(startTime int64, allocations storage.Storage[*allocat
 // This allows the handling of startups and re-placements of allocations.
 func handlePendingAllocationEvent(alloc *nomadApi.Allocation,
 	allocations storage.Storage[*allocationData], callbacks *AllocationProcessoring) {
-	log.WithField("alloc_id", alloc.ID).Debug("Handle Pending Allocation Event")
 	if alloc.DesiredStatus == structs.AllocDesiredStatusRun {
 		allocData, ok := allocations.Get(alloc.ID)
 		if ok && allocData.allocClientStatus != structs.AllocClientStatusRunning {
 			// Pending Allocation is already stored.
-			log.WithField("alloc_id", alloc.ID).Debug("Pending Allocation already stored")
+			// This happens because depending on the startup duration of the runner, we get zero, one, or more events
+			// notifying us that the allocation is still pending.
+			// We are just interested in the first event, in order to have the correct start time.
 			return
 		} else if ok {
 			// Handle Runner (/Container) re-allocations.
@@ -385,7 +383,6 @@ func handlePendingAllocationEvent(alloc *nomadApi.Allocation,
 // handleRunningAllocationEvent calls the passed AllocationProcessor filtering similar events.
 func handleRunningAllocationEvent(alloc *nomadApi.Allocation,
 	allocations storage.Storage[*allocationData], callbacks *AllocationProcessoring) {
-	log.WithField("alloc_id", alloc.ID).Debug("Handle Running Allocation Event")
 	if alloc.DesiredStatus == structs.AllocDesiredStatusRun {
 		// is first event that marks the transition between pending and running?
 		if allocData, ok := allocations.Get(alloc.ID); ok && allocData.allocClientStatus == structs.AllocClientStatusPending {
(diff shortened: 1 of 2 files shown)


internal/nomad/nomad.go
```diff
@@ -318,9 +318,6 @@ func handleEvaluationEvent(evaluations map[string]chan error, event *nomadApi.Ev
 // storage the state is persisted between multiple calls of this function.
 func handleAllocationEvent(startTime int64, allocations storage.Storage[*allocationData],
 	event *nomadApi.Event, callbacks *AllocationProcessoring) error {
-	if event.Type != structs.TypeAllocationUpdated {
```
MrSerth, line 321: Would it make sense to set another filter for the `PlanResult` here, rather than removing it?

internal/nomad/nomad.go
```diff
@@ -359,12 +356,11 @@ func handleAllocationEvent(startTime int64, allocations storage.Storage[*allocat
 // This allows the handling of startups and re-placements of allocations.
 func handlePendingAllocationEvent(alloc *nomadApi.Allocation,
 	allocations storage.Storage[*allocationData], callbacks *AllocationProcessoring) {
-	log.WithField("alloc_id", alloc.ID).Debug("Handle Pending Allocation Event")
 	if alloc.DesiredStatus == structs.AllocDesiredStatusRun {
 		allocData, ok := allocations.Get(alloc.ID)
 		if ok && allocData.allocClientStatus != structs.AllocClientStatusRunning {
 			// Pending Allocation is already stored.
-			log.WithField("alloc_id", alloc.ID).Debug("Pending Allocation already stored")
+			log.WithField("alloc_id", alloc.ID).Trace("Pending Allocation already stored")
```
MrSerth, line 363: Do you expect that (normally) we see this Trace message being logged? Just by reading the code and your explanation, I would argue that we first process the `PlanResult` and then the `AllocationUpdated` (thus causing this message)

MrSerth (commented):
Nice, I feel we are getting closer to identify the root cause! 💪 

Would you mind adding more details to the commit message, to make clearer why we just remove three lines in the code if we miss events? This aspect is currently not that clear, I feel (but I got it, thanks to the ticket you linked).

internal/nomad/nomad.go
mpass99: So far in all Influx events, `PlanResult` and `AllocationUpdated` are the only types of events within the `Allocation` topic.

But yeah, let's add the condition with the two topics, and create a Sentry warning in case of a new noticed type?!

internal/nomad/nomad.go
mpass99: Yes, I would agree that this log would be triggered for almost every started allocation. In order to not blur the Poseidon log, I degraded it to the `trace` level that is commonly not activated. But, I can also remove it?!

internal/nomad/nomad.go
MrSerth: I was rather thinking of _just_ filtering for `PlanResult`. This would also resolve the issue of duplicated log statements (see the other comment). However, if you say that both events are useful, I would rather follow your suggestion.

internal/nomad/nomad.go
MrSerth: At least, I think the message could be misleading, because it somewhat implies that the processing an already pending allocation is somewhat unexpected. In case we keep it (see above), I would rather opt to remove the line and expand the comment above (stating that we get two events per pending allocation and -- by design -- are just interested in a set of pending allocation, implying to remove duplicated columns).

internal/nomad/nomad.go
mpass99: Ah, I see. At this point, we cannot filter for just `PlanResult` as we depend on the event `Allocation`, `AllocationUpdated`, `running` (ClientStatus).

But, we could in line 361 (see the other comment). I don't know if that would be useful because with our current condition we are deciding on the state of our data, otherwise we would decide on the believe that just the `pending` `AllocationUpdated` event should be filtered out..

MrSerth: Mh, if you are saying "we depend on the event Allocation, AllocationUpdated, running (ClientStatus)", how this it work then? This somewhat implies that listening to the `PlanResult` does not provide all information?

mpass99: Right. Our process is split depending on the Client Status

https://github.com/openHPI/poseidon/blob/f377b1376c59a4f3f2a95258678e917e9fef69c5/internal/nomad/nomad.go#L343-L347

We need to handle the `pending` case in order to measure the startup time, and to handle container re-allocations.
We need to handle the `running` case in order to add just ready runners to the idle runners.

For both Client status, we observed 1 to 4 events [[1]](https://github.com/openHPI/poseidon/files/11462924/different_number_of_nomad_events.log). Therefore, we need to filter for just the first event each.

We cannot use only one type (`AllocationUpdated`, `PlanResult`) as we have many `pending` allocations with just the `PlanResult` event, and a couple of `pending` allocations with just the `AllocationUpdated` event [[2]](https://github.com/openHPI/poseidon/files/11463115/one_type_of_pending_events.txt).

MrSerth: Okay, interesting. This somewhat sounds like we need to combine both events, in order to get a complete picture? Probably, this supports your proposal to switch to the `Job` events rather than the `Allocation` events.

Let's give it a try and see what happens :)

MrSerth (approved):

PR #374, opened 2023-05-12
Debug/#358 log nomad events

mpass99:
We assume that influxdb cannot deal with the "time" tag key and deletes the whole batch of data containing this tag key.


MrSerth (approved):
Wow, this was unexpected. I am happy you noticed the issue that quickly, as we would have lost many events otherwise.

Your changes look right to me and since we were able to confirm that they ~~fix~~ circumvent the issue, I would just suggest merging and deploying. 🪂

PR #380, opened 2023-05-21
Fix increased prewarming pool size

mpass99:
by checking the number of required runners before creating an additional runner.


MrSerth (approved):
That's a good log statement. Let's observe it; do we need some additional events to recover what happened?

mpass99:
> do we need some additional events to recover what happened?

The recent bug could be identified well with the current amount of logs. Also, with the new changes we are not loosing any events :ok_hand:

MrSerth:
Alright, thanks for the explanation!

PR #381, opened 2023-05-21
Fix increased prewarming pool size

mpass99:
that is caused by all rescheduled allocations being added as idle runners including the used ones.

### DIFF
--- a/internal/nomad/executor_api_mock.go
+++ b/internal/nomad/executor_api_mock.go
@@ -343,11 +343,11 @@ func (_m *ExecutorAPIMock) SetJobScale(jobID string, count uint, reason string) 
 }
 
 // WatchEventStream provides a mock function with given fields: ctx, callbacks
-func (_m *ExecutorAPIMock) WatchEventStream(ctx context.Context, callbacks *AllocationProcessoring) error {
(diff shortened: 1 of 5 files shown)


MrSerth (approved):
I am fine with deleting allocations in case they were rescheduled. Probably, it would also be okay to accept a temporary increase of runners, but just to be sure, let's actually remove them.

MrSerth:
In today's meeting, we decided not to merge this PR. It would offer a partial solution for an increased pre-warming pool size, as identified with https://github.com/openHPI/poseidon/issues/358. Specifically, it would remove those allocations that were restarted by Nomad even if used already.

This removal is, however, not required: As soon as an instance is restarted, all user-generated data is also removed implicitly and thus the allocation is returned to a well-known, pre-defined state. Thus, those restarted allocations can be considered new and used for any upcoming request.

The only advantage of the removal would be to limit the number of runners waiting in the pre-warming pool to be as close as possible to the desired number. Still, this PR would only consider those allocations that were previously used (and, for example, were OOM killed). Other allocation restarts, e.g., as caused by the maintenance of Nomad agents, would not be handled by this PR.

Therefore, this PR would only take care of a few allocations restarts, which effects are already mitigated through our mechanism implemented with https://github.com/openHPI/poseidon/pull/380. Besides that, it would add further complexity with a limited use, thus we decided against it.

PR #383, opened 2023-05-28
Manually handle allocation restart

mpass99:
Refactor Allocation restart to move the responsibility of restarting allocations from Nomad to Poseidon.

The main focus of this PR is to handle restarts of the Docker service. The restart leads to all container being stopped. With this PR we stop Nomad from restarting these containers in order to do it by ourselves. We want this control because we only want to restart the idle runners and not already used runners.

We restart only every failed, not-used allocation of environments that has idle runners (/ at least one successful runner). This catches allocation failures due to the job specification (such as inaccessible Docker image).
Other causes of failed allocation will be handled by the `util.RetryExponential` functionality of the `Sample` limiting the number of retries to 18.

Related to #358
Replaces #381

### DIFF
--- a/internal/runner/nomad_manager.go
+++ b/internal/runner/nomad_manager.go
@@ -133,7 +133,7 @@ func (m *NomadRunnerManager) keepRunnersSynced(ctx context.Context) {
 	retries := 0
 	for ctx.Err() == nil {
 		err := m.apiClient.WatchEventStream(ctx,
-			&nomad.AllocationProcessoring{OnNew: m.onAllocationAdded, OnDeleted: m.onAllocationStopped})
+			&nomad.AllocationProcessing{OnNew: m.onAllocationAdded, OnDeleted: m.onAllocationStopped})
 		retries += 1
 		log.WithContext(ctx).WithError(err).Errorf("Stopped updating the runners! Retry %v", retries)
 		<-time.After(time.Second)
@@ -177,7 +177,7 @@ func monitorAllocationStartupDuration(startup time.Duration, runnerID string, en
 	monitoring.WriteInfluxPoint(p)
 }
 
-func (m *NomadRunnerManager) onAllocationStopped(alloc *nomadApi.Allocation) {
+func (m *NomadRunnerManager) onAllocationStopped(alloc *nomadApi.Allocation, failed bool) {
 	log.WithField("id", alloc.JobID).Debug("Runner stopped")
 
 	if nomad.IsEnvironmentTemplateID(alloc.JobID) {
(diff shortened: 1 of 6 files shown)


internal/runner/nomad_manager.go
```diff
@@ -190,9 +190,21 @@ func (m *NomadRunnerManager) onAllocationStopped(alloc *nomadApi.Allocation) {
...
 	m.usedRunners.Delete(alloc.JobID)
+
 	environment, ok := m.environments.Get(environmentID.ToString())
 	if ok {
 		environment.DeleteRunner(alloc.JobID)
+
+		if !isUsed && failed {
+			entry := log.WithField("failed_job", alloc.JobID)
+			if runner, ok := environment.Sample(); ok {
+				environment.AddRunner(runner)
```
MrSerth, line 203: What is the reasoning behind sampling and re-adding the runner just sampled? If I read the code correctly, you get a clean (unused) runner first, just to re-add it to the list of available unused runners the line after. Is this just to trigger the re-creation of a new runner?

I assume it is probably related to your statement:
> We restart only every failed, not-used allocation of environments that has idle runners (/ at least one successful runner). This catches allocation failures due to the job specification (such as inaccessible Docker image).

In any way, if we decide to keep this system, we should probably add a suitable test case here (since this will mechanism will be required in production).

MrSerth (commented):
Thanks for pointing out an alternative to the problem and taking care of restarts. I am not quite sure yet how to proceed here. For sure, having a more fine granular control over the restart process is beneficial to ensure we are always in charge of the runners. However, as you already pointed out, Nomad is perfect in rescheduling runners and ensuring they run as expected. If, in this scenario, we miss a few events, we will be even more off than usual.

MrSerth:
In today's meeting, we decided not to merge this PR. It would offer a partial solution for an increased pre-warming pool size, as identified with #358. Specifically, it would move the responsibility to handle allocation restarts from Nomad to Poseidon. This mechanism of having Poseidon handle these restart decisions might, however, be more error-prone than leaving it with Nomad (which is built to take care of this). Sure, it would allow us to prevent the restart of an allocation that got stopped (i.e., due to an OOM kill), but at the cost that all restarts would need proper handling. 

Currently, we don't think that this shift in responsibility is required, and we fear that we would miss too many edge cases (in which the pre-warming pool size would be too small). Since our overall goal is to ensure that the pre-warming pool size is not reduced over time, this mechanism would threaten said goal. Furthermore, allocations are rarely restarted and if they are, our mechanism implemented with #380 helps to return to the desired number of runners.

PR #390, opened 2023-06-03
Filter lost allocation events

mpass99:
Related to #357


internal/nomad/nomad.go
```diff
@@ -376,7 +379,11 @@ func handlePendingAllocationEvent(alloc *nomadApi.Allocation,
 			start:             time.Now(),
 			allocNomadNode:    alloc.NodeName,
 		})
-	} else {
+	case structs.AllocDesiredStatusStop:
+		// We delete the local data of this allocation / runner as we will likely receive event re-creating it.
```
MrSerth, line 383:
```suggestion
		// We delete the local data of this allocation / runner as we will likely receive an event re-creating it. If not, the next restart of Poseidon will re-add it.
```

MrSerth: Does this imply that we are also recreating a new runner, in case the pre-warming pool decreased below the desired value?

MrSerth (commented):
Thanks! Your code looks fine (what about a test?), but I am still unsure with regard to ticket #357. Sure, you're preventing POSEIDON-2E, but since the events in production (and potentially staging) are very diverse, I am not yet sure how to proceed. Will this change improve the situation in case of a lost Nomad agent? 

Otherwise, regarding your concern for a permanent list: I thin, we don't have a really good monitoring for that yet. Sure, Ansible will fail when the host is not reachable at all. However, we won't notice it immediately, if a Nomad service just fails silently. Let's extract this issue, since it is somewhat independent from the current one.

internal/nomad/nomad.go
mpass99: No. As you stated, a decreased pre-warming pool is currently refilled only on the next restart of Poseidon.

mpass99:
> what about a test?

Good point! I will add some soon.

> Will this change improve the situation in case of a lost Nomad agent?

**What is the problem again with the case of a lost Nomad agent?**

Our biggest concern is about too few runners. _But, shouldn't Nomad restart those runners in case of a lost allocation?_ Although, [[1]](https://discuss.hashicorp.com/t/understanding-job-restart-behaviour-on-lost-jobs/39353/2) connotes the idea of `lost` jobs being restarted, [the official documentation](https://developer.hashicorp.com/nomad/docs/job-specification/reschedule) just states that `failed` allocations are being restarted.
In our test we could identify `Lost Allocations that passed the runner timeout and are correctly not rescheduled`[[2]](https://codeocean.sentry.io/issues/4206026379/events/ea80d626421b482fa0f5fe9516499853/?project=6077585) and [`Lost allocations being rescheduled`](https://codeocean.sentry.io/issues/4206026379/events/aedc96a28d7347be83cb47abcde26644/?project=6077585&referrer=next-event). Accordingly, we did not had too few runners after it (in our production test) (See `2023-05-31 23:42:00`).

So, our problem is about the Sentry issues that created the impression that we have an issue to care about?!

**Which factors influence if an Nomad event should be reported to Sentry?**

1. The `ClientStatus` as we do not want normal events to be reported (`pending`, `running`, `complete`)
2. The `NextAllocation` flag as we do not want abnormal events that are being handled by Nomad to be reported
3. The `DesiredStatus` flag as we do not want lost allocations being stopped anyway to be reported [[2]](https://codeocean.sentry.io/issues/4206026379/events/ea80d626421b482fa0f5fe9516499853/?project=6077585)

With [139c10e](https://github.com/openHPI/poseidon/pull/390/commits/139c10e6e5acd3ddf28bc322a7dc0d6138fddef3) we refactor the event handling to follow a common structure and handle these factors. **Currently WiP, I will review it and write tests for it soon**

In future work, we might detect the need to not only report, but also handle the reported events. For this case we would also have to find identifier to differentiate i.e. between jobs that should be restarted and jobs that shouldn't (i.e. due to a invalid Docker image).

> Otherwise, regarding your concern for a permanent list: Let's extract this issue

#392

PR #396, opened 2023-06-14
Fix decreased prewarming pool due to inactivity timer

mpass99:
Related to #358 

Currently we observe some runners being `OOM Killed` (just) in the Poseidon logs (Lack of Error reporting or ok?). This leads to the runner being restarted and added to the idle runners again (even if it was used before the restart). In this case [POSEIDON-2N ](https://codeocean.sentry.io/issues/4220082532)`Too many idle runner` is expected for each restarted and re-added runner.

In this process we could identify two weaknesses:

1. We observed that the WebSocket connections for in that moment running executions are not directly stopped but ran into the execution timeout (of 30 seconds). In these 30 seconds much could happen. I.e. the runner gets restarted, re-added to the idle runners, and re-assigned to another user. When the WebSocket connection the got cancelled (due to the timeout), CodeOcean removes this runner. But, this runner now already belongs to another user.
Worst Case: The used runner gets removed and the other user has to request again a runner for execution.
I haven't found a concrete case for this scenario but the found cases convince me of the possibility.
2. Another case we have found is the case of runner `29-4dbd13a6-0acd-11ee-a4f7-fa163e079f19` that we fix with this PR. This case points out the bug that the inactivity timer is not stopped when the runner is restarted on a failure. The restart adds the runner to the prewarming pool again, but when the not-stopped timer expires the idle runner is removed again.

### DIFF
--- a/internal/runner/nomad_manager.go
+++ b/internal/runner/nomad_manager.go
@@ -190,8 +190,11 @@ func (m *NomadRunnerManager) onAllocationStopped(runnerID string) (alreadyRemove
 		return false
 	}
 
-	_, stillActive := m.usedRunners.Get(runnerID)
-	m.usedRunners.Delete(runnerID)
+	r, stillActive := m.usedRunners.Get(runnerID)
+	if stillActive {
+		r.StopTimeout()
+		m.usedRunners.Delete(runnerID)
+	}
 
 	environment, ok := m.environments.Get(environmentID.ToString())
 	if ok {
--- a/internal/runner/nomad_manager_test.go
+++ b/internal/runner/nomad_manager_test.go
@@ -367,6 +367,42 @@ func (s *ManagerTestSuite) TestOnAllocationAdded() {
 	})
 }
 
+func (s *ManagerTestSuite) TestOnAllocationStopped() {
+	s.Run("stops inactivity timer", func() {
+		testStoppedInactivityTimer(s, true)
+	})
+	s.Run("stops inactivity timer - counter check", func() {
+		testStoppedInactivityTimer(s, false)
+	})
+}
+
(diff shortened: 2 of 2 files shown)


MrSerth (approved):
That's a very interesting observation -- thanks for detailing that.

I know that we might see an `OOM killed` message, when too much memory is consumed within a runner. One example to trigger this error might be the following: In an R script, just type the following:

```r
table(airquality)
```

This will create a table overview of one of the internal data sets called `airquality`. It exceeds the RAM limit we defined, so it will fail and will be killed. I am not sure, how this is persisted in logs, but it is a case that might happen (and that is properly handled by CodeOcean). Here is a list of the most recent occurrences (in UTC) that occurred _after_ June 4th as recognized by CodeOcean:

```
2023-06-14 16:42:47.078677
2023-06-14 11:46:20.098001
2023-06-14 07:03:31.789496
2023-06-13 14:58:46.524764
2023-06-12 10:17:47.363491
2023-06-10 21:57:59.818506
2023-06-10 17:57:28.871130
2023-06-09 16:07:47.630225
2023-06-08 07:17:15.938918
2023-06-08 07:17:15.921321
2023-06-07 07:29:30.087396
2023-06-07 07:25:06.726970
2023-06-06 12:11:35.386035
2023-06-06 06:21:23.736571
2023-06-05 19:58:38.049195
```

1. If this memory limit we defined is the root cause of those issues, the WebSocket connection to the browser should also be closed, but of course I don't know about the other WebSocket connection between Poseidon and CodeOcean. I have the feeling that in any case, we should also stop the WebSocket connection if a runner becomes unavailable (such in this case). Do you see any possibility to tackle this issue as well?
2. The second case you describe is logical and the fix makes sense, thanks so much!

PR #407, opened 2023-07-14
Refactor/#316 sentry tags

### LINKED ISSUE #316: Sentry Additional Data Deprication
The Additional Data Dictionary that we use for error events to provide information about the error (environment id, error message) is [deprecated](https://docs.sentry.io/platforms/javascript/enriching-events/context/#additional-data).

Instead, we are advised to use a "Context" to store the data.

Furthermore, sometimes it would be useful to search the events for information that is in the additional data. But, the additional data (and the context?) is not searchable. We could think about adding the information (about the environment id) as "Tag".

mpass99:
Closes #316


MrSerth (approved):
Nice, thanks for generalizing some parts (and removing those duplications, e.g., by moving the `monitoring.AddRunnerMonitoringData` to the middleware :+1:

PR #417, opened 2023-08-01
Fix/#325 incomplete debug message

### LINKED ISSUE #325: Exec debug message could not be read completely
Sentry issue: [POSEIDON-38](https://codeocean.sentry.io/issues/4336294196/?referrer=github_integration) (former: [POSEIDON-1X](https://codeocean.sentry.io/issues/3941992590/?referrer=github_integration))

```
Exec debug message could not be read completely
```

Some spans, mostly those about `unset`, don't contain a time information and thus are invalid.

mpass99:
Closes #325 

In order to fix the incomplete debug message we aim to ignore output of the runner after we have sent the `SIGQUIT`.

We could do so by:
- moving the `SentryDebugWriter` away from the debug command creation (in `nomad.go`) to the event processing (in `nomad_runner.go`).
- Let the `SentryDebugWriter` probe if it can write to the `CodeOceanWriter` before it processes the data.
- Passing a second context from `nomad.go` to the `nomad_runner.go` that is done when no output should be transferred (before the execution should be aborted).

I liked the second option the most which is now implemented in this PR.
The regression test is in a separate commit as it took more changes than the actual fix.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -127,6 +127,8 @@ jobs:
       POSEIDON_AWS_ENABLED: false
       POSEIDON_AWS_ENDPOINT: ${{ secrets.POSEIDON_AWS_ENDPOINT }}
       POSEIDON_AWS_FUNCTIONS: ""
+      POSEIDON_LOGGER_FORMATTER: "JSONFormatter"
+      POSEIDON_LOG_FILE: "../../poseidon.log"
       POSEIDON_NOMAD_DISABLEFORCEPULL: true
       GOCOVERDIR: coverage
     steps:
@@ -204,7 +206,7 @@ jobs:
           until curl -s --fail http://localhost:4646/v1/agent/health ; do sleep 1; done
           chmod +x ./poseidon
           mkdir -p ${GOCOVERDIR}
-          ./poseidon &
+          ./poseidon | tee poseidon.log &
           until curl -s --fail http://localhost:7200/api/v1/health ; do sleep 1; done
           make e2e-test
       - name: Run e2e recovery tests
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -230,7 +230,7 @@ func main() {
 	if err := config.InitConfig(); err != nil {
 		log.WithError(err).Warn("Could not initialize configuration")
 	}
-	logging.InitializeLogging(config.Config.Logger.Level)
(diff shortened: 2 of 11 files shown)


MrSerth:
Awesome, this PR (and the knowledge gained to create it) is great news! Many thanks as well to separate the regression test in a dedicated commit and writing it first (that's real TDD)!

Your code changes look fine, and I recognize the probing you suggest here. There is just a minor thing I wasn't sure about: When & how is determined that the `CodeOceanWriter` isn't writable any longer? Is it through [this context](https://github.com/openhpi/poseidon/blob/eb818f92f760f0e9e7a29d3e0a38fbbc78c10dea/internal/api/ws/codeocean_writer.go#L107-L108), that gets changed in this PR?

From my (limited) understanding, the second, or third option sound good, and I am fine to go with the second one you implemented. :+1:

mpass99:
> When & how is determined that the `CodeOceanWriter` isn't writable any longer? Is it through [this context](https://github.com/openhpi/poseidon/blob/eb818f92f760f0e9e7a29d3e0a38fbbc78c10dea/internal/api/ws/codeocean_writer.go#L107-L108), that gets changed in this PR?

Exactly, this context represents if the `CodeOceanWriter` is writable or not.
- When the client disconnects, the context is canceled here:
https://github.com/openHPI/poseidon/blob/ed1b83d13cda903109b84f1490b65474a4e60657/internal/api/websocket.go#L50-L52
  - Additionally, we cleanly stop the goroutine here:
  https://github.com/openHPI/poseidon/blob/8490e34d6a82feb7faa087ad35cfc1289a9859f1/internal/api/websocket.go#L65-L68
- When an (external) error appears when sending WebSocket messages, the context is cancelled (indirectly) here as `done` will be `true`:
https://github.com/openHPI/poseidon/blob/ed1b83d13cda903109b84f1490b65474a4e60657/internal/api/ws/codeocean_writer.go#L135-L137
- When the execution is done (the normal case), we `Close` the `CodeOceanWriter` here:
https://github.com/openHPI/poseidon/blob/8490e34d6a82feb7faa087ad35cfc1289a9859f1/internal/api/websocket.go#L72-L75
  - This leads to the context being cancelled (indirectly) here as `message.done` will be `true`:
  https://github.com/openHPI/poseidon/blob/ed1b83d13cda903109b84f1490b65474a4e60657/internal/api/ws/codeocean_writer.go#L135-L137

MrSerth (approved):
Awesome, thanks for your detailed explanation. This makes sense to me, and I was now able to fully understand it :)

PR #419, opened 2023-08-14
Update Go to 1.21

MrSerth:
This PR updates to the latest Go version 1.21. I recommend having a read of the [official changelog](https://go.dev/doc/go1.21) for the new Go version.

- There is one notice about Go tools version (regarding backwards compatibility), but this shouldn't affect us.
- The Profile-guide optimization (PGO) is now enabled by default (with the `-pgo=auto` switch). I checked whether this allows us to get rid of some of our manual action in the Makefile, but don't think so: In order to distinguish between PGO and non-PGO builds, we inject a variable value there. This is still required:
https://github.com/openHPI/poseidon/blob/90092c48c164ef3269b8405a6db9d4a0ed33f799/Makefile#L55


mpass99 (approved):
Great :+1: While reading the release notes I noticed the additional context functionality. I think it would suit well for the two changed occurrences

MrSerth (commented):
> I think it would suit well for the two changed occurrences

✅ Perfect, that's an awesome addition -- thank you! 👍

PR #422, opened 2023-08-21
Fix/#406 missing log

mpass99:
Related to #406


MrSerth (approved):
This will make debugging so much easier, I like that! 👏

Issue #429, opened 2023-08-26
Out-Of-Memory on Poseidon VM


MrSerth:
Recently, I observed a variety of different errors with Poseidon, including [CODEOCEAN-T1](https://codeocean.sentry.io/issues/4421478383/), [CODEOCEAN-S9](https://codeocean.sentry.io/issues/4412338743/), [CODEOCEAN-S8](https://codeocean.sentry.io/issues/4412335958/), [CODEOCEAN-SR](https://codeocean.sentry.io/issues/4417388132/).

Those errors all appeared since Monday evening, where I performed a (manual) deployment of commit c0a3fb12c3d48ca2e3a8d59a5be1477b521a31e3 to our server. While first considering network issues, I now think we have a problem with our Poseidon VM (not necessarily with Poseidon, but I don't know yet). My first investigation showed: We have a memory issue!

Here is the overall system memory to that VM:

<img width="1706" alt="Bildschirmfoto 2023-08-26 um 16 41 20" src="https://github.com/openHPI/poseidon/assets/7300329/0e80a992-f155-4da8-889e-3b9a98416bb3">

Further, [the kernel log](https://github.com/openHPI/poseidon/files/12445596/oom.log) clearly shows that the OOM killer terminates Poseidon. We should investigate here immediately to restore regular operation.

MrSerth:
Indeed, Poseidon seems to be responsible for the majority of memory:

<img width="1077" alt="Bildschirmfoto 2023-08-26 um 18 03 46" src="https://github.com/openHPI/poseidon/assets/7300329/aec331fc-2244-4005-baf0-55785a074520">

Shortly after, this already reads as

<img width="1077" alt="Bildschirmfoto 2023-08-26 um 18 21 51" src="https://github.com/openHPI/poseidon/assets/7300329/368247e3-3ae6-46e0-bc20-c2a3d3da75b1">

mpass99:
With #430 we fix this specific issue.

Anyway, I think we should monitor goroutine leaks in general, i.e. by adding tests that count and compare the number of goroutines before and after. What do you think?

MrSerth:
Thanks for fixing the issue (and deploying it already)!

Sure, if there is a method to check for leaked go routines, I'd go with that. Not sure how easy such a test is to achieve, but if you have an idea, I am very open to it. :+1:

mpass99:
ToDo: Also check for shuffling the test order

MrSerth:
[...]

PR #430, opened 2023-08-26
Fix Memory Leak caused by the merge_context.

mpass99:
Related to #429; Already deployed

The now removed statement of sending an empty struct into the channel blocked the goroutine until the channel of Done got listened for. This led to a goroutine leak as one does not necessarily has to call the Done function of a context.

We fix this issue by removing this value. It was unnecessary anyway as a closed channel always returns the null-value of the returned type.


MrSerth (approved):
Awesome, thanks for fixing the issue that fast! Even though I am not an expert in the channels, this seems to be a reasonable change for me. Let's give it a try!

PR #431, opened 2023-08-28
Introduce Memory Leak checks

### LINKED ISSUE #429: Out-Of-Memory on Poseidon VM

mpass99:
by adding an assertion about the number of Goroutines to the unit tests.

Closes #429

This is a time-boxed start of introducing memory checks to the unit tests. This requires for each test the confrontation if the increased number of Goroutines is justified.

In the recent small time-box only the first three test files are refactored. As next step, I would continue with the remaining test files which would take some time (but not too much)..


MrSerth (commented):
Seems good so far, thank you. I just have one minor question regarding the general test setup: How independent are the tests? For some parts, I've seen you increased the Goroutine counter with the same comment multiple times in a file (which is okay). But aren't you otherwise reusing the same environment, so that an increase (e.g., for the monitored idle runner storage) is not needed?

mpass99:
> But aren't you otherwise reusing the same environment, so that an increase (e.g., for the monitored idle runner storage) is not needed?

Yes, it depends on the test itself if objects (runner manager, monitored storages, ...) are reused. In the tests so far it got not reused. In other tests, such as the `nomad_manager_test.go` the `NomadRunnerManager` is already created in the `SetupTest` and therefore somehow reused. In this case we would add the `s.ExpectedGoroutingIncrease++` just next to it in the `SetupTest`. Did I understand your question correctly?

MrSerth:
> Did I understand your question correctly?

Yes, you did. Okay, your explanation makes sense. Still, the test setup is not that easy for me to follow, because in our Ruby on Rails projects each test is independent of any other (which is not completely the case here).

mpass99:
Yeah, sub-tests with a test can be dependent, tests within a test suit, and of course e2e tests. Anyway, it was a good hint as I had the exact error with not resetting the `ExpectedGoroutingIncrease` before [d5696fb](https://github.com/openHPI/poseidon/pull/431/commits/d5696fb934c53d804d4dceeccb20f91ec2db00c6) :ok_hand:

MrSerth:
I just converted your previous comment with the bullet points into a to-do list, so that we can track was has been done already. Since the decided to ignore one issue, I already ticked that.

MrSerth (approved):
Looks good! What about the tree open issues, are they done?

mpass99:
Yes, they are all done.
1. [90ec1a4](https://github.com/openHPI/poseidon/pull/431/commits/90ec1a489be5bcd1bbc0b8282d2d6068ce65f344)
2. [5f8e990](https://github.com/openHPI/poseidon/pull/431/commits/5f8e990acac294811433db9a7413ad6741c2065f)
3. -
4. [b39d334](https://github.com/openHPI/poseidon/pull/431/commits/b39d334dc5de0a520a2b2430135c5495cb9e187f)

MrSerth:
> Yes, they are all done.

Perfect 👌

PR #433, opened 2023-08-29
Today's lost runners

mpass99:
Today at 2 am UTC we have experienced another drop of idle runners (likely connected with the unattended upgrade).
We have [the Nomad events](https://github.com/openHPI/poseidon/files/12467698/eventsOfLostRunner.txt) and [the Poseidon logs](https://github.com/openHPI/poseidon/files/12467719/PoseidonLogs.txt) regarding one example runner. The Nomad events are clear: We have an allocation that gets lost and rescheduled into another allocation belonging to the same job. The influxdb shows that Poseidon has not registered the rescheduled allocation - pointing at the bug that we drop all rescheduled idle runners because the Nomad Manager does not check if the deleted runner was still active in the prewarming pool.


internal/runner/nomad_manager_test.go
```diff
@@ -373,6 +374,16 @@ func (s *ManagerTestSuite) TestOnAllocationAdded() {
 }
 
 func (s *ManagerTestSuite) TestOnAllocationStopped() {
+	s.Run("returns false for idle runner", func() {
```
MrSerth, line 377: Shouldn't we test the case that `true` is returned (i.e., a runner gets deleted twice, or the runner was not found) as well?

MrSerth (approved):
Awesome, many thanks for acting proactively here and observing the errors! 💯 

I just have a minor remark regarding the tests, but that's not too important. Also, I tried following the issue with the two logs you provided, but found it a little difficult: Especially the Nomad log is very verbose with many details per line, making it harder to identify the relevant parts. It worked out in the end, but still took me some time to dig through the logs (and finally get a hint with the next / prev. allocations; also using the search function for the description / client status / desired status / ... turned out to be useful).

internal/runner/nomad_manager_test.go
mpass99: I added a test for the scenario that a runner was not found.
This test simulates the race condition behavior of the inactivity timer (with the Nomad event stream handling) that led to the requirement of returning this boolean. The inactivity timer removes the runner (shortly before an allocation rescheduling event was processed).

MrSerth: Awesome, thanks!

MrSerth (changes requested):
Nice, thank you! 🦄

PR #454, opened 2023-09-17
Fix the recent event of lost runners

mpass99:
In the recent host migration, different Nomad hosts got restarted. This PR handles the case of runner `29-6f04b525-5315-11ee-af32-fa163e079f19`. What happens in its [Poseidon Log](https://github.com/openHPI/poseidon/files/12643314/poseidonLog.txt) is documented in the test case `TestHandleAllocationEvent_RegressionTest_14_09_2023`.

In summary, multiple consecutive reschedulings led again (such as in #433) to the wrong triggering of the race condition handling. As this handling (introduced in f031219cb8cd02f289abb4d107396fd7457a69f6) already caused multiple bugs and its purpose seems to be just to suppress incorrect warnings in case of a race condition, I replaced it with an implementation that does not influence the process.

### DIFF
--- a/internal/nomad/nomad.go
+++ b/internal/nomad/nomad.go
@@ -453,7 +453,8 @@ func handlePendingAllocationEvent(alloc *nomadApi.Allocation, allocData *allocat
 			stopExpected:       stopExpected,
 		})
 	case structs.AllocDesiredStatusStop:
-		handleStoppingAllocationEvent(alloc, allocations, callbacks, false)
+		// As this allocation was still pending, we don't have to propagate its deletion.
+		allocations.Delete(alloc.ID)
 		// Anyway, we want to monitor the occurrences.
 		if !allocData.stopExpected {
 			log.WithField("alloc", alloc).Warn("Pending allocation was stopped unexpectedly")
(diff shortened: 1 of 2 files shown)


MrSerth (approved):
Awesome, looks good! Many thanks also for writing a specific test first actually reproducing the events we saw. I am sure this will help us to prevent another occurrence of the same type in the future.

I am also fine to monitor the desired stopped events. However, I am just wondering: Is the desired status `stop` only for those runners stopped through a migration? I thought it is issued for _each_ runner being stopped, also for those that are requested by Poseidon. In the latter case, we'll probably see a lot of those messages... (Anyway, I am fine to get a better impression of the occurrence rate, so feel free to keep it).

When reviewing the commits, I just noticed a small typo in the commit message of the second commit for "pending".

mpass99:
> Is the desired status `stop` only for those runners stopped through a migration? I thought it is issued for _each_ runner being stopped, also for those that are requested by Poseidon.

Yes, you are right the `DesiredStatus` `stop` is not only for runners stopped through a migration, but rather quite normal. But, in this function, we only handle pending allocations (`ClientStatus` `pending`) and the occurrence of a pending allocation being stopped is uncommon. To get a feeling of how uncommon this is, I introduced that debug (warn-level) log statement.

MrSerth:
Alright, thanks for clarifying!

PR #455, opened 2023-09-17
Reuse rescheduled allocations

mpass99:
As we've already discussed, we want to reuse rescheduled (formerly used) runners, as they are in a clean state. While we do so for `OOM Killed` allocations (see 6a1677dea090e62815e830dd08dcef9baf494669), we did not for other rescheduled allocations. [The case of runner](https://github.com/openHPI/poseidon/files/12643430/poseidonLog.txt) `29-f5d671f7-5316-11ee-af32-fa163e079f19` made this clear.


MrSerth (approved):
Awesome, glad we took the previous allocation reschedules (introduced by the VM migration) to resolve these issues as well :+1:

PR #456, opened 2023-09-19
Fix WebSocket JSON schema

mpass99:
by changing the `required` attribute to be an array as described in [the official documentation](https://json-schema.org/understanding-json-schema/reference/object.html#required).


MrSerth (approved):
Looks good and works with JSONSchemer in CodeOcean as well.

Issue #453, opened 2023-09-17
Another Out-Of-Memory Error on Poseidon


MrSerth:
We are experiencing another out-of-memory error on Poseidon, that should be investigated further.

I don't have many details to share, yet. The issues appeared first on Sunday, September 17th with release 39fc0f9. After rolling back to the previous commit 68cd8f43b451ac39ad55c272881f3db66f401156 (around 3:30pm CEST), the issue did not occur another time. This could point to one of the dependencies, but I am not sure about that either.

<img width="1713" alt="Bildschirmfoto 2023-09-17 um 15 40 43" src="https://github.com/openHPI/poseidon/assets/7300329/d092b790-63a9-4dd6-a90e-a86e27201add">

MrSerth:
We have another occurrence of an OOM kill just from this morning (Sep 18 06:38:45). See more details with `cat /var/log/syslog | grep -A 50 -B 50 OOM`.

mpass99:
Unfortunately, Poseidon's Log does not really help identify the memory leak. During the OOM kill, three actions were happening:
1. One running execution (next to 60338 others that day)
2. One runner stopped due to an inactivity timer (which happened also 5136 other times this day without problems)
3. One allocation event handling (23944 times that day)

Accordingly, I was also not able to reproduce the issue in the staging environment.
#457 might help identify the leak.

MrSerth:
> Unfortunately, Poseidon's Log does not really help identify the memory leak. 

I feared that, but thanks for looking into it nevertheless. Besides the PR #457 you've just created, do we have a chance to analyze a core dump or (performance) profile? I am still hoping it would be possible to get a memory dump just before the memory is cleared, which might further assist us in debugging the cause.

MrSerth:
Ah, thanks for investigating here! I just went ahead and disabled the profile sampling (by setting `profilessamplerate: 0.0`). With this, we should be able to confirm our assumption about the root cause. If this doesn't help, we could also consider downgrading the sentry dependency to limit the cause there.

MrSerth:
It seems like the issue didn't occur another time, that's good! 🥳 

[...]

PR #457, opened 2023-09-19
Introduce debug memory monitoring

mpass99:
that alerts and prints additional debug information in case Poseidon exceeds 1 GB of memory usage.

Related to #453


MrSerth (approved):
I am fine to add such a debug statement and hope that the memory increase is "slow enough" to trigger it.

PR #477, opened 2023-10-06
Formalize Memory Monitoring

### LINKED ISSUE #452: Send (OOM) kills to Sentry
Currently, when Poseidon is killed, e.g., by the OOM killer, we do not get notified nor any Sentry event. Ideally, we would add some hook or other mechanism to get informed whenever Poseidon quits unexpectedly.

mpass99:
Closes #452 

by extracting the interval and threshold into configuration options.

Related to f670b07e.

### DIFF
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -85,8 +85,8 @@ func shutdownSentry() {
 }
 
 func initProfiling(options config.Profiling) (cancel func()) {
-	if options.Enabled {
-		profile, err := os.Create(options.File)
+	if options.CPUEnabled {
+		profile, err := os.Create(options.CPUFile)
 		if err != nil {
 			log.WithError(err).Error("Error while opening the profile file")
 		}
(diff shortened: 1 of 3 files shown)


cmd/poseidon/main.go
```diff
@@ -112,18 +112,19 @@ func initProfiling(options config.Profiling) (cancel func()) {
...
-	// We assume that Poseidon usually takes about 50-300 MB of memory. Therefore, we specify the threshold of 1 GB.
-	// Improve: Make this value dynamic or relative.
-	const threshold = 1 * 1000 * 1000 * 1000
-	const interval = 5 * time.Second
+func watchMemoryAndAlert(options config.Profiling) {
+	if options.MemoryInterval == 0 {
+		return
+	}
 
+	var exceeded bool
```
MrSerth, line 120: When is the value of this variable set? I see two usages below, but none of them changes the value. Am I missing something?

mpass99: No, I missed adjusting its value. Thank you for catching that!

MrSerth: Ah, okay. Now, it makes sense. You remember whether the last check already saw an exceeded memory use and each time this observation changes, you issue a corresponding message.

MrSerth (approved):

PR #481, opened 2023-10-12
Add error for empty list file system execution.

mpass99:
Related to #436 

Normally, the result of executing the `lsCommand` should never be empty. However, we have observed that CodeOcean sometimes receives an empty JSON result if the runner is being deleted while the list file system request is processed. Therefore, we add a check if something has been written to CodeOcean and otherwise report an error.

This check does not fix this issue but confirms whether we deal with an empty result of the Nomad execution.


MrSerth (approved):
Sure, let's give it a try 👀

mpass99:
Neither locally nor in three further CI runs, I could reproduce the error in the e2e tests of the first and second CI-run.
Instead, the CI is failing because of some lint-preparing-error.
Therefore, I will continue merging this PR.

Issue #470, opened 2023-10-04
Lost Runners, again


MrSerth:
On October 4th, we lost some runners again, leading to an empty pre-warming pool and errors being displayed to our learners.

Timestamps shown in CEST:
<img width="467" alt="Bildschirmfoto 2023-10-04 um 08 38 48" src="https://github.com/openHPI/poseidon/assets/7300329/fdb8ca6d-e21b-4cf8-860c-c405f581e7a4">

MrSerth:
I just had to resync the runners again, since we lost some another time.

mpass99:
The runners were lost on Oct 4 from 7:44 to 7:53.

Context:
- Poseidon was restarted at 7:43 UTC
  - Have you done a deployment then?
- The recovery of runners and allocations lasts until 7:50
- Within minute 7:44 several file copy requests to Nomad failed, but the Nomad event stream continued. After that, the requests succeeded.
- At 7:50 the Nomad event stream times out and is restarted.
- At 7:53 the first runner is created again.
- Between the Poseidon restart and 7:53
  - every requested runner led to one less idle runner that will not be replaced.
  - The Nomad server logs contain issues reaching the Nomad server 3 `nomad.stats_fetcher: failed retrieving server health: server=nomad-server-terraform-3.global error="context deadline exceeded"`
  - The Nomad agents do not report issues and contain logs about running executions
  - Poseidon has received events from Nomad notifying that some allocations are now running, but no events about pending allocations.

All in all, I would argue that Nomad has lost some requests in the state that its health pings to the Nomad server 3 timed out (maybe due to the difference of requests timing out vs requests not getting accepted :shrug:). As we have too little information about this and an issue within Poseidon is not revealed, I would accept (/ignore) this case and investigate the next occurrences.

Possibilities how to address this case would be:
1. Log every Evaluation (update) via Influxdb. Evaluations are the tracking options of Nomad. For each longer-running request (such as creating a Job) the HTTP request is returned immediately with an evaluation ID. Via the event stream, further status updates for the evaluation are provided.
2. Introduce alerting if an evaluation is not closed within a timespan (of 10 minutes).
3. Introduce alerting if an evaluation does not lead to (a pending and then) running job/allocation within a timespan.
4. Expand the retry mechanism also to recreate the runner if it is not created within the timespan.

MrSerth:
Thanks for digging deep on the first occurrence.

> Poseidon was restarted at 7:43 UTC
> - Have you done a deployment then?

[...]

Issue #265, opened 2022-12-28
Disclosure of internal error messages.


mpass99:
For some routes such as `listFileSystem`, we send internal error messages (for example received from Nomad) to the client.

This is [a common problem](https://owasp.org/www-community/Improper_Error_Handling) that got popular for exploiting SQL Injection Vulnerabilities.

We should:
- send only a few specific/know/controlled error messages to the client if required.
- think about where to log the internal error messages (in the `api` package?)
- think about how useful the `HTTPLoggingMiddleware` is as it does not contain specific debug information

MrSerth:
Thanks for pointing that out! Actually, I really like the error messages for debugging, but I agree that we probably should limit those for our production environment.

MrSerth:
We agreed on a two-way approach for sensitive error messages:

- Specific details about an issue shall not be sent through the API, but rather logged in Poseidon or sent to Sentry
- The API should mainly return useful error messages, that do not contain internal error details (such as those received from Nomad).

Additionally, we would like to slightly modify the error handling with Sentry. Currently, we always sent two events: One regarding the failed HTTP request (e.g., `Failing /api/v1/runners`) and one with the more specific error (e.g., `Error reading client message`). In order to make Sentry more useful, we would like to focus on those specific error messages and remove the more generic ones about failing HTTP request. If we need statistical data about HTTP requests and the return code, we can use Influx and Grafana for that.

PR #482, opened 2023-10-23
Recover Runners on Event Stream Restart

mpass99:
Related to #470 

I tried to keep the changes small, but the codebase was only adjusted for only one recovery of the environments and runners. This led to multiple issues that environments and runners are replaced "dirty" - without destroying the environment and runner references leading to leaked goroutines.


MrSerth:
Nice done, thanks! I think, the code looks fine, so that I am focussing on the functionality in my review.

As I've noticed, you've deployed the code to staging, which allowed me to test it there. I did so using a Ruby exercise containing the following program:

```rb
30.times do |index|
  puts index
  sleep 1
end
```

Then, I went ahead and restarted all Nomad Servers through OpenStack Horizon, definitely causing the Event Stream to break:

Bildschirmfoto 2023-10-29 um 17 57 55

This action triggered two actions:
1. [CODEOCEAN-JP](https://codeocean.sentry.io/issues/3977036924/) was triggered
2. All runner data was deleted locally in Poseidon (besides some more [log entries](https://github.com/openHPI/poseidon/files/13198673/poseidon.log))

However, even after the Nomad servers were back online and the cluster working again, none of the runners were recovered (all execution environments were reported as having 0 idle runners). Shouldn't this be resolved with this PR, so that the test I performed is handled correctly by Poseidon? Further, I was wondering whether we should add some test cases to the CI for some behaviour, too?

mpass99:
First of all, thank you for your testing! As you have noticed, I've deployed the code to staging and also tested this scenario by restarting the Nomad service. It worked fine then. But, identifying that an error case remains when rebooting the servers is very important!

> [CODEOCEAN-JP](https://codeocean.sentry.io/issues/3977036924/) was triggered

I would argue it's valid that CodeOcean receives an error when the requested operation (`list_filesystem`) has an error (in this case Nomad was unreachable). For me, the question arose whether Poseidon should disclose the whole internal error message. In this case, it was good for debugging. We may discuss this in the context of #265 

> All runner data was deleted locally in Poseidon (besides some more [log entries](https://github.com/openHPI/poseidon/files/13198673/poseidon.log))

Till this point, the logs look valid.

> Even after the Nomad servers were back online and the cluster working again, none of the runners were recovered

The difference in your scenario is that the request timed out. The i/o timeout error is a `context.DeadlineExceeded` error - the same error we use for checking if the retry mechanism should continue. We fix this behavior in commit [111cc0f](https://github.com/openHPI/poseidon/pull/482/commits/111cc0fa9a0ec2363c17ef041c8f32eb1a1474b0).

 <details>
  <summary>Side story</summary>
  Poseidon does not always behave the same for the same error.

``` log
Oct 30 16:24:09 poseidon-terraform poseidon[39546]: time="2023-10-30T16:24:09.906289Z" level=warning msg="Loading Environments failed! Retrying..." error="couldn't load template jobs: couldn't load jobs: error listing Nomad jobs: Get \"https://nomad.internal-codemoon.xopic.de:4646/v1/jobs?namespace=poseidon&prefix=template\": dial tcp 10.224.6.99:4646: i/o timeout" package=environment
Oct 30 16:24:10 poseidon-terraform poseidon[39546]: time="2023-10-30T16:24:10.907828Z" level=debug msg="retrying after error" count=1 error="couldn't load template jobs: couldn't load jobs: error listing Nomad jobs: Get \"https://nomad.internal-codemoon.xopic.de:4646/v1/jobs?namespace=poseidon&prefix=template\": dial tcp 10.224.6.99:4646: i/o timeout" package=util
```

``` log
Oct 30 16:27:57 poseidon-terraform poseidon[39546]: time="2023-10-30T16:27:57.356690Z" level=warning msg="Loading Environments failed! Retrying..." error="couldn't load template jobs: couldn't load jobs: error listing Nomad jobs: Get \"https://nomad.internal-codemoon.xopic.de:4646/v1/jobs?namespace=poseidon&prefix=template\": dial tcp 10.224.6.20:4646: i/o timeout" package=environment
Oct 30 16:27:58 poseidon-terraform poseidon[39546]: time="2023-10-30T16:27:58.359378Z" level=info msg="Stopped KeepEnvironmentsSynced" error="couldn't load template jobs: couldn't load jobs: error listing Nomad jobs: Get \"https://nomad.internal-codemoon.xopic.de:4646/v1/jobs?namespace=poseidon&prefix=template\": dial tcp 10.224.6.20:4646: i/o timeout" package=environment
```

In the first version, Poseidon retries after the timeout error. In the second version Poseidon stops retrying at all. But, the error looks the same.

The difference lies in the Golang packages. The `poll` package defines the `DeadlineExceededError` returning `i/o timeout`. It is used i.e. from the `net` package. The `net` package defines the `timeoutError` also returning `i/o timeout`, but it also defines that `timeoutError == context.DeadlineExceeded`. This case triggers the stop of our retry mechanism.

</details> 

> add some test cases to the CI for some behaviour, too?

Good point! In [464b496](https://github.com/openHPI/poseidon/pull/482/commits/464b4965974c632aae74ddbb0f4e0faab0776ab6) I've added four test cases. Can you think of more?

MrSerth (approved):
Awesome, the improved functionality and the test added look really nice! I've given the new changes another test run on Staging and can confirm that even longer testing periods without any Nomad server are handled properly (besides reporting many events to sentry, for each retry...).

Further, the tests are plausible and, in my opinion, cover the most relevant parts of the code! 👏

PR #498, opened 2023-11-17
Remove misleading warning.

### LINKED ISSUE #432: Nomad Event Stream failed when stopping Poseidon
While investigating another issue, I noticed that Poseidon regularly faces an issue when being stopped. In my examples, I was starting Poseidon, kept it running for 10-15 seconds (with Nomad working as expected) and then stopped it again. In this scenario, I often (but not always) saw an error:

```
time="2023-08-28T22:40:40.827735Z" level=debug msg="Creating new runners" environment_id=10 package=environment runnersRequired=0
time="2023-08-28T22:40:40.828015Z" level=info msg="Starting server" address="127.0.0.1:7200" package=main
^C
time="2023-08-28T22:40:54.199813Z" level=info msg="Received SIGINT, shutting down ..." package=main
time="2023-08-28T22:40:54.199948Z" level=info msg="Server closed" error="http: Server closed" package=main
time="2023-08-28T22:40:54.200008Z" level=debug msg="Stopping CPU profiler" package=main
[Sentry] 2023/08/28 22:40:54 Event dropped due to noopTransport usage.
time="2023-08-28T22:40:54.200025Z" level=error msg="Nomad Event Stream failed! Retrying..." count=1 error="<nil>" package=runner
```

I am wondering about the last error "Nomad Event Stream failed!" What failed here (especially since the `error` is `nil`) and is this behavior expected?

mpass99:
Closes #432

### DIFF
--- a/internal/environment/nomad_manager.go
+++ b/internal/environment/nomad_manager.go
@@ -180,7 +180,7 @@ func (m *NomadEnvironmentManager) KeepEnvironmentsSynced(synchronizeRunners func
 		}
 
 		// Load Runners and keep them synchronized.
-		if err := synchronizeRunners(ctx); err != nil {
+		if err := synchronizeRunners(ctx); err != nil && ctx.Err() == nil {
 			log.WithContext(ctx).WithError(err).Warn("Loading and synchronizing Runners failed! Retrying...")
 			return err
 		}


MrSerth (approved):
What a simple solution :+1:

I was testing this branch locally multiple times. There are two observations I made:

- When the CPU profiler is disabled (which I refer to as a combination of generating the profile and generating a profile-optimised build), the original issue never appeared in about 20-25 test cases. 
- When the CPU profiler is enabled, your changes reliably prevent the error.

I am still surprised by my observation of the CPU profiler settings (specifically `profiling.cpuenabled`), but am happy that stopping Poseidon is now as smooth as starting 😁

mpass99:
> I am still surprised by my observation of the CPU profiler settings

I would guess, we may have a race condition here.
``` go
	case <-shutdownSignals:
		log.Info("Received SIGINT, shutting down...")

		defer stopProfiling()
		ctx, cancel := context.WithTimeout(context.Background(), gracefulShutdownWait)
		defer cancel()
		if err := server.Shutdown(ctx); err != nil {
```

With `server.Shutdown` we close the server which leads (at the very end of this function) to the cancellation of the main context (which asynchronously leads to the closing of the Nomad Event Stream and the former error message). However, how much time this asynchronous handling has depends on `profiling.cpuenabled`:
- if not `profiling.cpuenabled`: Poseidon just `cancelInflux` before exiting.
- if `profiling.cpuenabled`: Poseidon `cancelInflux` and `stopProfiling` before existing. Especially `stopProfiling` can take quite long when `SIGUSR1` has not been sent before (as we then write all the profiling data before exiting).

Anyway, I don't see much incentive to fix this race condition as it just controls if we skip some (in this case irrelevant) error handling. Furthermore, I don't see good options to fix it. A simple but unreliable would be to add `runtime.Gosched` or `time.After` before exiting Poseidon. More effective, but complicated would be to observe if all (relevant) goroutines finished before existing.

MrSerth:
Ah, thanks for investigating and describing the potential race condition.

In this case, I agree that fixing probably isn't worth the effort and that the changes done in this PR are reducing the impact of the race condition. Of course, a more reliable solution would be great, but I also agree that it's probably not worth the effort. Therefore, I am fine with the current solution. :+1:

PR #507, opened 2023-11-23
Inject Execution Debug Message

### LINKED ISSUE #442: Sentry: Add Debug Message
We have no performance instrumentation for the `until [ -e /tmp/stderr_1693522759468167218.fifo ];` part of our command. This leads to the duration of the previous part being higher. Usually this manipulates the value of the setUser span.

We should fix this by introducing another debug message directly before the `until` part.

mpass99:
for measuring the performance of the until loop of the stderr connection.

Closes #442

Accidentally, I pushed this commit directly to `main`.  
In response, I reverted that commit on `main` and enabled the branch protection rule `Do not allow bypassing the above settings`.


MrSerth (approved):
Many thanks for adding the debug statement and enabling the rule. Otherwise, just adding this one line would have been fine for me as well ;)

PR #510, opened 2023-11-29
Adjust Influxdb buffering

### LINKED ISSUE #479: Improve error handling for sending events to Influxdb
Recently, we lost some monitoring data (see #471 and #470) that were not sent to Influxdb. Besides improving the monitoring for system services (#471), we are also interested in understanding why we didn't see a Sentry event for Poseidon failing to send data to Influxdb.

Therefore, this ticket is about checking the Influxdb error handling in Poseidon.

mpass99:
as we have experienced silent package drops. This issue is not fixed, it is just made less probable.

Closes #479

### DIFF
--- a/pkg/monitoring/influxdb2_middleware.go
+++ b/pkg/monitoring/influxdb2_middleware.go
@@ -57,18 +57,21 @@ func InitializeInfluxDB(db *config.InfluxDB) (cancel func()) {
 		return func() {}
 	}
 
+	// How often to retry to write data.
+	const maxRetries = 50
 	// How long to wait before retrying to write data.
-	const retryInterval = 30 * time.Second
+	const retryInterval = 5 * time.Second
 	// How old the data can be before we stop retrying to write it. Should be larger than maxRetries * retryInterval.
 	const retryExpire = 10 * time.Minute
-	// How often to retry to write data.
-	const maxRetries = 10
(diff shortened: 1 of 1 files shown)


pkg/monitoring/influxdb2_middleware.go
```diff
@@ -57,18 +57,21 @@ func InitializeInfluxDB(db *config.InfluxDB) (cancel func()) {
 		return func() {}
 	}
 
+	// How often to retry to write data.
+	const maxRetries = 10
```
MrSerth, line 61: Do you think we should increase the `maxRetries` as well? I am assuming that in the old version, we tried for a maximum of 30s * 10 = 300s = 5m, where we only retry 5s * 10 = 50s in the new version? Or are there some other implications I am not aware of?

MrSerth: I see, that's indeed true -- the thin line is not easy to hit. Nevertheless, I think that the increased `maxRetries` is a good thing: Sometimes, an instance might just become unavailable for various reasons (such as a reboot). In such a case, 50 seconds should normally be enough to transmit data (even though I haven't measured the reboot times), but the margin would be somewhat small. Therefore, increasing it a little makes sense to me. I think, we should just try it with the new values and observe the difference in Sentry.

PR #516, opened 2023-12-05
Systemd watchdog

### LINKED ISSUE #478: Systemd Watchdog
As a DevOps
I want to configure a systemd watchdog
in order to register unexpected downtimes of Poseidon.

- Poseidon should periodically send data to the socket provided by the environment variable.
- If the watchdog registers a downtime the service will be restarted.
- The watchdog should be able to handle service stops and deployments without false positives.

mpass99:
Closes #478 

Note: Even though the issue only aims to monitor the "Liveness" of Poseidon, we also introduced a "Readiness" check by using `systemd.service.Type=notify` such as in CodeOcean.

ToDo:
- [x] Think about a useful test design
- [x] Test on Staging

References: [[1]](http://0pointer.de/blog/projects/watchdog.html) [[2]](https://github.com/coreos/go-systemd/blob/main/daemon/watchdog.go) [[3]](https://vincent.bernat.ch/en/blog/2017-systemd-golang)


MrSerth:
Looks good so far! Signalling both, liveness and readiness is awesome -- happy that you include them.

cmd/poseidon/main.go
```diff
@@ -214,6 +218,67 @@
...
+}
+
+func notifyWatchdog(ctx context.Context, router *mux.Router, interval time.Duration) {
+	healthRoute, err := router.Get(api.HealthPath).URL()
+	if err != nil {
+		log.WithError(err).Error("Failed to parse Health route")
+		return
+	}
+	// We do not verify the certificate as we (intend to) perform only requests to the local server.
+	tlsConfig := &tls.Config{InsecureSkipVerify: true} // #nosec G402 The default min tls version is secure.
```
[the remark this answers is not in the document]
mpass99: I'm quite unsure about this. We might either dismiss this alert as the request is only local or add a Poseidon configuration option to pass the ca certificate and use it for this request. What do you think?

MrSerth (commented):
Looks good so far, thanks! Below ist the answer for the TLS verification as you requested.

Further, I wondered a little about the test coverage: Besides your e2e test and the `notify` activation in the minimal service, Codecov doesn't show the relevant lines (in method `notifyWatchdog`) to be covered: https://app.codecov.io/gh/openHPI/poseidon/pull/516

I already tried rerunning the pipeline (to upload code coverage again), but that didn't change here. Do you have an idea what could be missing?

configuration.example.yaml
```diff
@@ -19,6 +19,8 @@ server:
     # certfile: ./poseidon.crt
     # The path to the key file used for TLS
     # keyfile: ./poseidon.key
+    # The path to the ca certificate used for TLS requests.
+    # cafile: ./ca.crt
```
MrSerth, line 23: I would recommend to keep the same order as we have for `nomad.tls`. There, the `cafile` proceeds the `certfile` and `keyfile`.

configuration.example.yaml
```diff
@@ -19,6 +19,8 @@ server:
     # certfile: ./poseidon.crt
     # The path to the key file used for TLS
     # keyfile: ./poseidon.key
+    # The path to the ca certificate used for TLS requests.
```
MrSerth, line 22:
```suggestion
    # The path to the certificate of the CA authority used for TLS
```

MrSerth (commented):
Nicely done, thanks! I think, the code looks fine so far. I am currently waiting for the pipeline to proceed with the deployment on staging, so that I can test your changes with systemd. Otherwise, I am still wondering about the code coverage; did you get any insights here already?

MrSerth:
I just verified the changes on our staging environment and was able to verify them successfully. For my testing, I used `iptables` to block the request:

```
iptables -A INPUT -i lo -p tcp --destination-port 7200 -j DROP
```

As a consequence, systemd restarted the service as expected, showing that the watchdog worked 🥳.
(Dropping all iptable routes is done with `iptables -F`)

One final observation I made are the log entries. For each health check, we see the corresponding request in the systemd log:

```
time="2024-01-17T13:07:48.267735Z" level=debug code=204 duration="64.693µs" method=GET path=/api/v1/health user_agent=Go-http-client/1.1
```

Here, the `user_agent` `Go-http-client/1.1` hints at Poseidon requesting the health route as part of the watchdog. Do you think we should somehow modify that (e.g., by changing the user agent for this request [should be rather simple], lowering the statement to `trace` [rather difficult, if the change should only apply to this route])? Otherwise, I am also fine to leave it as it is, just wanted to raise the question.

mpass99:
> I am still wondering about the code coverage; did you get any insights here already?

Good catch that the coverage of the recovery tests is missing!
The issue was that we killed Poseidon instead of stopping it properly. Go writes the coverage only when the program exits (1), not when it is killed (2). So, we have to stop Poseidon properly by `SIGTERM`, instead of killing it with `SIGKILL`.
Also, when Poseidon is started as a service it does not automatically have the environment defined in the workflow. Therefore, we have to set the `GOCOVERDIR` explicitly (3).

mpass99:
> Do you think we should somehow modify that

Using our `HTTPLoggingMiddleware`, I think it would be simple to apply such filters. What would you like to filter exactly?
1. All Health requests
2. All Health requests with the user agent `Go-http-client/1.1`
3. All Heath requests with a statically defined token placed in the user agent
4. All Heath requests with a dynamically (or in the configuration) defined token placed in the user agent

I guess with the first three options, adversaries would be able to hide their Health requests in the logs...

MrSerth:
> What would you like to filter exactly?

Good question! 😅 

I would rule out the first option. Those health requests originating from an external system, even if just our own monitoring with Icinga, probably should still be included. My rational behind filtering _here_ would be that those requests "watchdog-triggered" are rather "internal" in a sense that the Poseidon binary on the very same system is making them. 

> I guess with the first three options, adversaries would be able to hide their Health requests in the logs...

Yes, that's what I thought as well. Probably, the way to resolve that would be to generate a secure string on application startup, use that for those local watchdog requests and filter only those requests containing this token. Currently, I don't see any other good reason to filter HTTP logs originating from any other system.

.github/workflows/ci.yml
```diff
@@ -216,13 +216,22 @@ jobs:
...
+      - name: Print Systemd Failure logs
         run: |
-          tail -f /var/log/syslog &
-          make e2e-test-recovery
+          /usr/bin/systemctl --user show poseidon.service -p NRestarts
+          journalctl --user -xe -u poseidon.service --no-pager
+        if: failure()
+      - name: Print Coverage files
+        run: |
+          killall poseidon
```
MrSerth, line 233: Wouldn't it make sense to stop Poseidon through systemd?

```suggestion
          systemctl --user stop poseidon.service poseidon.socket
```

I think, using `killall` is fine as well (that uses SIGTERM by default), but the other one would probably be similar to the actual usage (we are, nevertheless, still in an e2e context).

.github/workflows/ci.yml
```diff
@@ -216,13 +216,22 @@ jobs:
...
+        run: make e2e-test-recovery
+        if: ${{ success() || failure() }}
+      - name: Print Systemd Failure logs
         run: |
-          tail -f /var/log/syslog &
-          make e2e-test-recovery
+          /usr/bin/systemctl --user show poseidon.service -p NRestarts
+          journalctl --user -xe -u poseidon.service --no-pager
+        if: failure()
+      - name: Print Coverage files
```
MrSerth, line 231: This name doesn't seem to fit the purpose completely. Otherwise, I wouldn't expect that a print statement stops my application 😅

.github/workflows/resources/poseidon-minimal.service
```diff
@@ -8,5 +8,11 @@ Requires=poseidon.socket
 [Service]
 WorkingDirectory=${GITHUB_WORKSPACE}
 ExecStart=${GITHUB_WORKSPACE}/poseidon
-Restart=always
 Environment="POSEIDON_SERVER_SYSTEMDSOCKETACTIVATION=TRUE"
+Environment="GOCOVERDIR=${GITHUB_WORKSPACE}/${GOCOVERDIR}"
```
MrSerth, line 12: What do you think about extracting this variable `GOCOVERDIR` and the other ones `.github/workflows/ci.yml` (lines 126-132) to a dedicated file? 

Then, we could use this file in the systemd service with `EnvironmentFile=<Path>` and in the remaining parts with one of the solutions sketched [here](https://stackoverflow.com/questions/19331497/set-environment-variables-from-file-of-key-value-pairs)?

My main goal would be to reduce duplication of the variables. Otherwise, I would feat that we miss adjusting the variables again in the future.

MrSerth (commented):
> > I am still wondering about the code coverage; did you get any insights here already?
> 
> Good catch that the coverage of the recovery tests is missing! The issue was that we killed Poseidon instead of stopping it properly. Go writes the coverage only when the program exits (1), not when it is killed (2). So, we have to stop Poseidon properly by `SIGTERM`, instead of killing it with `SIGKILL`. Also, when Poseidon is started as a service it does not automatically have the environment defined in the workflow. Therefore, we have to set the `GOCOVERDIR` explicitly (3).

Awesome, many thanks for digging into this topic and taking the time to discover and resolve those issues. While being able to follow your explanation, it would have caused me hours to discover that.

One final question on the coverage file: s ensured that the coverage file is _appended_ when we run the recovery tests (after previously finishing the "actual" e2e tests? Or are two files created automatically? I just want to ensure that we are not overwriting the "previous" results by ensuring a proper exit of the recovery test service...

mpass99:
> Is it ensured that the coverage file is _appended_ when we run the recovery tests (after previously finishing the "actual" e2e tests? Or are two files created automatically?

For each start of Poseidon (/the binary) Go creates a separate coverage file (next to a shared file with meta information). Therefore, the coverage of multiple Poseidon restarts (such as with the e2e tests and the recovery tests) is merged.

MrSerth:
Let's use the following user agents:

`Poseidon/<7 Git Hash> Go-http-client/1.1` -- external requests
`Poseidon/<7 Git Hash> (XXX) Go-http-client/1.1` -- Watchdog

pkg/dto/dto.go
```diff
@@ -9,6 +9,17 @@ import (
 	"strings"
 )
 
+var (
+	// UserAgentOut for outgoing requests (without libraries). The Git Hash will be replaced by main.go.
+	UserAgentOut      = "Poseidon/" + UserAgentVCSPlaceholder + " Go-http-client/1.1"
```
mpass99, line 14: Currently, we only have the Watchdog doing requests without a library. I'm unsure if it's worth the complexity of teaching the libraries how to use our User Agents while we currently have no benefit from doing so 🤷

MrSerth: Alright, fine for me. I assume that all Nomad requests are handled by their library anyway, so that's fine :+1:.

MrSerth (approved):
Awesome, that's really a great work and I am looking forward to see this feature in action! I was just making a minor adjustment to the description of our newest addition in the config, and will proceed merging.

PR #542, opened 2024-01-26
Debug Health route latency.

mpass99:


MrSerth (approved):
Looks good, thanks for adding some debug statements. I'll proceed merging and deploying the changes, so that we can observe their effect.

While testing your changes on staging, I observed that not all cases might be covered by the debug statement: Only those requests that arrive and are handled by the router can be tracked. If, for some reason, the outgoing request would be delayed before reaching the router / handler, this won't be logged. That said, I would still prefer to give your changes a try and just observe what happens :+1:

Issue #451, opened 2023-09-11
Unit Test Shuffling


mpass99:
- [ ] Check if our unit tests are executed within a random order
- [ ] Enable test shuffling [[1]](https://dev.to/gkampitakis/go-test-supports-shuffling-tests-38e)
- [ ] Fix broken test dependencies

PR #543, opened 2024-01-26
Feature/#478 systemd watchdog

mpass99:
Implements "Issue" 0 of https://github.com/openHPI/poseidon/issues/541#issuecomment-1912579294

PR #544, opened 2024-01-26
Decouple InfluxDB writings from request handling.

mpass99:
With #451, we found that writing an InfluxDB data point might block and lead to high latencies.

Implements "Issue" 1 of https://github.com/openHPI/poseidon/issues/541#issuecomment-1912579294


MrSerth (approved):
Thanks! What a simple and straightforward solution :+1:.

In order to get the desired results of our investigation introduced with #543, I would actually wait with merging and deploying until the next occurrence of the watchdog-triggered restart occurs. As soon as that happened, I am fine testing this change to confirm that no further restarts are caused by the watchdog.

PR #556, opened 2024-02-27
Fix linter emptyStringTest rule

mpass99:
by replacing the length check with a string comparison. This rule was introduced with the new GolangCI lint version.


MrSerth (approved):
Awesome, thanks!

PR #557, opened 2024-02-27
Update Golang to 1.22

mpass99:
It is considering the last Update of Golang (2f43bce) and the changes that make the separate handling in the GitHub CI obsolete (39fc0f9d).


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/openHPI/poseidon
 
-go 1.21.1
+go 1.22
```
MrSerth, line 3: I was trying to identify why I switched to `1.21.1` in a previous commit. This was somewhat related to some tool. If I extracted that correctly from my browser's history, I was motivated by the following comment:

```
https://github.com/golang/go/issues/62278#issuecomment-1693538776
```

Has that behavior changed? If I read the comments in the aforementioned issue correctly, that's not the case. However, I cannot reproduce the issue locally any longer (except by running `GOTOOLCHAIN="go1.22+auto" go mod tidy`, which I didn't do before -- I only executed `go mod tidy` or actually `make tidy-deps`).

go.mod
MrSerth, line 1: Would you mind taking another look at the dependencies, please?

When running `make upgrade-deps` on the current branch, I get:

```
go: downloading github.com/coreos/go-systemd v0.0.0-20191104093116-d3cd4ed1dbcf
go: downloading github.com/influxdata/influxdb-client-go v1.4.0
go: downloading github.com/go-jose/go-jose v2.6.2+incompatible
go: downloading github.com/hashicorp/consul v1.17.3
go: downloading github.com/hashicorp/go-secure-stdlib v0.1.0
go: downloading github.com/hashicorp/go-set v0.1.14
go: downloading github.com/hashicorp/vault v1.15.5
go: downloading github.com/apparentlymart/go-textseg v1.0.0
go: downloading github.com/armon/go-metrics v0.5.3
go: downloading github.com/shirou/gopsutil v3.21.11+incompatible
go: downloading github.com/gojuno/minimock v1.9.2
go: downloading github.com/apapsch/go-jsonmerge v1.0.0
go: downloading google.golang.org/genproto v0.0.0-20240221002015-b0ce06bbee7c
go: downloading github.com/cenkalti/backoff v2.2.1+incompatible
go: trying upgrade to github.com/armon/go-metrics@v0.5.3
go: github.com/armon/go-metrics@v0.5.3: parsing go.mod:
	module declares its path as: github.com/hashicorp/go-metrics
	        but was required as: github.com/armon/go-metrics
	trying github.com/armon/go-metrics@v0.5.2
go: github.com/armon/go-metrics@v0.5.2: parsing go.mod:
	module declares its path as: github.com/hashicorp/go-metrics
	        but was required as: github.com/armon/go-metrics
	trying github.com/armon/go-metrics@v0.5.1
go: github.com/armon/go-metrics@v0.5.1: parsing go.mod:
	module declares its path as: github.com/hashicorp/go-metrics
	        but was required as: github.com/armon/go-metrics
	trying github.com/armon/go-metrics@v0.5.0
go: github.com/armon/go-metrics@v0.5.0: parsing go.mod:
	module declares its path as: github.com/hashicorp/go-metrics
	        but was required as: github.com/armon/go-metrics
	trying github.com/armon/go-metrics@v0.4.2
go: github.com/armon/go-metrics@v0.4.2: parsing go.mod:
	module declares its path as: github.com/hashicorp/go-metrics
	        but was required as: github.com/armon/go-metrics
	restoring github.com/armon/go-metrics@v0.4.1
go: warning: github.com/hashicorp/go-msgpack@v1.1.5: retracted by module author: v1.1.5 merged upstream ugorji/go which breaks compatibility with previous versions.
go: to switch to the latest unretracted version, run:
	go get github.com/hashicorp/go-msgpack@latest

Process finished with exit code 0
```

Or is this something we cannot change, since the `go-metrics` module is required indirectly?

MrSerth, line 1: The list of (indirect) dependencies might contain some outdated dependencies, at least as my IDE highlights.

For example, the dependency `github.com/golang/protobuf` is outdated and can be replaced by `google.golang.org/protobuf`. Or is that something we cannot change either, since it is an indirect dependency?

MrSerth (commented):
Thanks for taking a look here. In general, the upgrade looks fine, but I still struggle to work with the Go dependency files (two recent questions detailed below).

go.mod
mpass99: I see, I wasn't aware of this. Now, I changed the version to `1.22.0`. That is also the fix for most of the linked pull requests.

go.mod
mpass99: Both, the dependency warnings for `github.com/golang/protobuf` and for the retracted `github.com/hashicorp/go-msgpack` version are imported due to our use of `github.com/hashicorp/nomad/nomad/structs`. We might be able to `replace` the dependency, but as we cannot be sure that we generate issues doing so, I would refrain from doing so.

mpass99: Following https://github.com/hashicorp/nomad/issues/11826 we can also include a `replace` statement to fix this error.

go.mod
MrSerth: Don't worry, I wasn't aware of that either and really had to focus on finding some plausible explanation for my change 🤓. Obviously, the commit message wasn't that great, back then 🙈 

Thanks for changing. Now, we also need to be aware of this specific version and increase it more often (as minor Go releases get published).

go.mod
MrSerth: Okay, I see. Obviously, we need to wait for upstream changes. So far, we only have one `replace` directive to resolve a security warning and I would suggest keeping that practice: We (mostly) rely on the default resolver, but modify it if real arguments promise an advantage (like resolving a security warning, or fixing some production-related error).

MrSerth: Well, as mentioned below, I would then probably just continue to wait and somewhat "accept" those messages out of necessity. The `go-metrics` module is, in that version, directly required by Nomad, which in turn is one of our main dependencies. I am not happy with these warnings, but have to admit that we probably don't have the resources (and at least I the knowledge) to catch up with another solution.

MrSerth (approved):

go.mod
mpass99: Alright, sounds reasonable 👍

MrSerth (approved):
Looks good now, ready to merge.

PR #566, opened 2024-03-28
Fix Runner DNS resolution

### LINKED ISSUE #490: Network access becomes unavailable
An execution environment providing network access first seems to work correctly. However, after some time (and some unknown events), the environment looses the network access. The allocation itself is still running on Nomad, but unfortunately without any possibility to reach the internet. 

Within a bash container, you can test network access through:

```bash
curl api.ipify.org
```

So far, we don't know yet when the error occurs. However, resynchronizing the environment from CodeOcean fixes the issue.

Bildschirmfoto 2023-11-01 um 18 52 43

mpass99:
Closes #490 

The failing DNS resolution was caused by Nomad overwriting the DNS configuration of Docker and the CNI secure bridge. To set valid Nameservers, we have to configure the according Nomad options.

### DIFF
--- a/configuration.example.yaml
+++ b/configuration.example.yaml
@@ -65,6 +65,14 @@ nomad:
   # namespace: poseidon
   # Prefer local Docker images over pulling them from a registry. Images with the `latest` tag will always be force pulled by Nomad regardless of this configuration.
   disableforcepull: true
+  # Network configuration for network-enabled runners. See https://developer.hashicorp.com/nomad/docs/job-specification/network.
+  network:
+    # Available Modes: "none", "bridge", "host", "cni/*".
+    # "none": Even the network-enabled runners will be isolated.
+    # "bridge": Isolated network namespace with bridged interface. Linux-only.
+    # "host": Using the host network namespace. Less-secure.
+    # "cni/*": Configure an isolated network namespace using CNI. For example, this can be a more secured bridge network.
+    mode: "cni/secure-bridge"
 
 aws:
   # Specifies whether AWS should be used as executor.
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -8,6 +8,7 @@ import (
 	"flag"
 	"fmt"
 	"github.com/getsentry/sentry-go"
+	nomadApi "github.com/hashicorp/nomad/api"
 	"github.com/openHPI/poseidon/pkg/dto"
 	"github.com/openHPI/poseidon/pkg/logging"
 	"github.com/sirupsen/logrus"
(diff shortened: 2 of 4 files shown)


MrSerth (commented):
Many thanks for working on this fix!

To be honest, I have mixed feelings about this PR and the changes. At first, I was very surprised by the approach to fix the secure bridge: I thought (and confirmed) that a fix is also possible by just changing the `secure-bridge.conflist`. The required changes are shown in the details.

<details><summary>Details</summary>
<p>


```json5
"type": "bridge",
"bridge":"nomad-filtered",
// More options as usual...
"dns":{
    "nameservers":[
        "8.8.8.8",
        "8.8.4.4",
        "2001:4860:4860::8888",
        "2001:4860:4860::8844",
    ],
    "domain": "poseidon.internal", // Probably, can be omitted. It didn't have any effect
    "search": [
        "poseidon.internal"
    ]
}             
```

</p>
</details> 

Can you please elaborate why you decided to use this mechanism or where you see the advantages / disadvantages?

I think, having everything in the `secure-bridge.conflist` might be beneficial to have everything in one place (Routing + DNS). Nevertheless, with the config option on Poseidon, we also get some further flexibility.

<hr>

On a second thought and playing around with this PR, I got more in favour of your changes. I clearly see the benefits to extract the network config to settings, especially on my local machine (finally allowing me to tweak the `nomad.network.mode` setting, so that I can use another value and get network-enabled instances. And it works even without the CNI plugins, which is also great. 

Maybe do both (this PR and the `secure-bridge.conflist`)? Or is that bad, since we would need to update something in two locations?

mpass99:
> Can you please elaborate why you decided to use this mechanism

Thanks for bringing that up! After double-checking, it seems the main reason was a deployment issue.  
I deployed a similar DNS configuration in the `secure-bridge.conflist`, but it didn't apply. Then I found the [Nomad documentation](https://developer.hashicorp.com/nomad/docs/job-specification/network#dns) defining exactly the observed behavior: "By default all task drivers will inherit DNS configuration from the client host. [...] if you are using a mode="cni/*, these values will override any DNS configuration the CNI plugins return". Therefore, the case was clear to me.

At this point, having both options available again, I would put the DNS configuration back to our Deployment repository underlining your argument that we have all the other Routing and DNS configurations also there.

> Maybe do both (this PR and the `secure-bridge.conflist`)? Or is that bad, since we would need to update something in two locations?

Let's use the `secure-bridge.conflist` to configure the DNS. If you like this "feature" we might keep it but only configure `nomad.network.mode` in our deployment configuration, so we use the DNS options defined in the `secure-bridge.conflist`.

MrSerth:
Alright, I agree with your proposal. And yes, I would actually like to continue having this feature, since it allows me to run the network part on my operating system, too :)

MrSerth (approved):

MrSerth (approved):
These code changes work like a charm, just tested these locally ✅

PR #578, opened 2024-04-25
Synchronize CNI secure bridge documentation

mpass99:
with production use.


MrSerth:
Nice, thanks! Shall we just add suitable routes for IPv6 too (i.e., running the script only with the static exclusions, not the Ansible hosts)?

MrSerth (approved):
Thanks!

PR #585, opened 2024-05-07
Fix Golangci-lint configuration

mpass99:
Open to discussion: By specifying the linters explicitly in an allow list, we are missing all the linters that are added in each release. Should we drop the allow list and introduce a deny list if necessary?

### DIFF
--- a/.golangci.yaml
+++ b/.golangci.yaml
@@ -1,3 +1,6 @@
+run:
+  go: "1.22"
+
 linters-settings:
   errcheck:
     check-type-assertions: true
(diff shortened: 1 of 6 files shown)


MrSerth (approved):
The code changes look good, thanks!

Regarding the configuration: I would prefer a deny list, since this allows us to benefit from new additions of the linter. Ideally, a dependabot would open a new PR with a new linter version, so that we can see whether the new version bump includes new checks that need manual intervention before the pipeline succeeds (all other updates that still have a succeeding pipeline are a no-brainer, I would say).

PR #586, opened 2024-05-07
[Lint] Change to linter deny list and fix formerly excluded linters

mpass99:


MrSerth:
From my point of view, this PR is currently on stale, besides the good intent (and our general desire to proceed with this topic). I added it to the project and would suggest to continue it as part of our current maintenance / bug fixing efforts.

.golangci.yaml
```diff
@@ -42,56 +43,13 @@ linters-settings:
...
   exclude-rules:
     - path: _test\.go
       linters:
-        - noctx
         # Always closing the HTTP body unnecessarily complicates the tests
         - bodyclose
+        - funlen
+        - noctx
         # We don't need to wrap errors in tests
         - wrapcheck
```
MrSerth, line 55: Would you mind adding an empty new line at the end of this file?

MrSerth (dismissed):
Many thanks for this tedious work -- I think it is a really huge improvement! 👏 

It's also great to see that you were creating so many commits, which made it rather easy for me to review the changes on a commit-by-commit bases. Also, the comments (especially when disabling a linter or on other decisions) were helpful. Overall, I reviewed all commits individually, and learned more about Golang linters during some web searchers 🙂

MrSerth (approved):

PR #611, opened 2024-06-12
Nomad Restart and Reschedule Policy

mpass99:
Related to #587 

Fix restart and reschedule configuration to not reschedule infinitely (e.g. due to an invalid image specifier).

ToDo:
- [ ] ~~Should we write an e2e test checking that a wrong environment definition does not loop infinitely? With the current configuration, the test will last almost 10 minutes..~~ No

### DIFF
--- a/internal/environment/template-environment-job.hcl
+++ b/internal/environment/template-environment-job.hcl
@@ -11,11 +11,6 @@ job "template-0" {
       sticky  = false
     }
     count = 1
-    scaling {
-      enabled = true
-      min = 0
-      max = 300
-    }
     spread {
       // see https://www.nomadproject.io/docs/job-specification/spread#even-spread-across-data-center
       // This spreads the load evenly amongst our nodes
--- a/internal/environment/template-environment-job.hcl
+++ b/internal/environment/template-environment-job.hcl
@@ -18,11 +18,18 @@ job "template-0" {
       weight = 100
     }
     restart {
-      delay = "0s"
+      attempts = 3
+      delay = "15s"
+      interval = "1h"
+      mode = "fail"
     }
     reschedule {
-      unlimited = true
-      attempts = 0
+      unlimited = false
+      attempts = 3
+      interval = "6h"
+      delay = "1m"
+      max_delay = "4m"
+      delay_function = "exponential"
     }
 
     task "default-task" {


internal/environment/template-environment-job.hcl
```diff
@@ -11,23 +11,24 @@ job "template-0" {
       sticky  = false
     }
     count = 1
-    scaling {
```
mpass99, line 14: This seems to be a historical artifact from the time we had multiple tasks per job. The [documentation](https://developer.hashicorp.com/nomad/docs/job-specification/scaling) does not point out a benefit for our current use.

internal/environment/template-environment-job.hcl
```diff
@@ -11,23 +11,24 @@ job "template-0" {
...
     spread {
       // see https://www.nomadproject.io/docs/job-specification/spread#even-spread-across-data-center
       // This spreads the load evenly amongst our nodes
       attribute = "${node.unique.name}"
       weight = 100
     }
     restart {
-      delay = "0s"
+      attempts = 3
+      delay = "15s"
```
mpass99, line 22: Is this delay still justified, even if the old delay of `0s` seems to be unrelated to the observed issue?

It might help in some cases which we have not observed yet. On the other hand, it adds additional delay in the provisioning of the (idle) runners.

internal/environment/template-environment-job.hcl
MrSerth: Thanks for identifying that, I am happy to remove it.

internal/environment/template-environment-job.hcl
MrSerth: Mh, I might not get your comment completely. Let's start from the last sentence:

> On the other hand, it adds additional delay in the provisioning of the (idle) runners.

Yes, but only if some error occurred. In the success case where everything runs smoothly, this should not make any difference at all, should it?

> It might help in some cases which we have not observed yet.

Maybe. But wasn't it designed to help with #587?

> Is this delay still justified, even if the old delay of `0s` seems to be unrelated to the observed issue?

Why is the old delay of `0s` unrelated to the observed issue? Is this related to the updated working theory mentioned in https://github.com/openHPI/poseidon/issues/587#issuecomment-2163662374? I currently feel there is some gap I am missing...

MrSerth (commented):
Thanks!

I would not add an e2e for this scenario. We would mainly test Nomad (and not something truly Poseidon specific), which doesn't feel right. Further the long increase is really not ideal. If at all, I would opt to add another template job with shorter wait times specifically for the given scenario. Still, I don't think it's the right thing to do...

internal/environment/template-environment-job.hcl
mpass99: > In the success case where everything runs smoothly, this should not make any difference at all, should it?

That's right 👍 And I think in this case Poseidon at some point returns an error to CodeOcean, resulting in the request for a new runner (so the delay really is asynchronous).

> But wasn't it designed to help with https://github.com/openHPI/poseidon/issues/587?
> Why is the old delay of 0s unrelated to the observed issue?
> Is this related to the updated working theory mentioned in https://github.com/openHPI/poseidon/issues/587#issuecomment-2163662374?

Exactly. The updated working theory is based on the evaluation of `are we currently rescheduling at all (unlimited) or not (attempts)?`. We there found that we are currently rescheduling unlimited. Further, we found that if Nomad has no agents eligible and tries to restart or reschedule a failed Allocation, it directly deregisters the job and does not try to restart or reschedule after the respective delays.

All in all, I'm fine with keeping the 15s delay.

MrSerth: Okay, thanks!

> Further, we found that if Nomad has no agents eligible and tries to restart or reschedule a failed Allocation, it directly deregisters the job and does not try to restart or reschedule after the respective delays.

Then, the delay of 15s would be helpful, since this could mean an agent could have become eligible in the meantime again? For example: A job is running, all agents go do for a simple service restart, waits x delay seconds, the server reschedules. With a delay of x = 0s, the job will fail permanently and will be deregistered. With a delay of x = 15s, one agent could be eligible again, couldn't it? Or would we need to increase x further (based on our Ansible config), so that it is longer than 15 (`nomad_service_stop_timeout`) + 5 (for systemd `TimeoutStopSec`) + (~5 or ~10) (for an estimated service start) = ~30s?

Alternatively, we could rely on Poseidon to handle this permanent job failure (in which case x might not be as important as otherwise assumed / calculated). However, I am not sure about the advantages and disadvantages of the handling in Poseidon vs. Nomad.

mpass99: Together, we identified the implications of Nomad restarts ("Connected Server", "Other Server", "All Server", "Allocation Agent Only", and "All Agents") on "idle runners", "currently streaming runners", and "used but inactive runners".

Regarding this conversation, we found that the `restart` policy is not decisive because on a deployment the Agent hosting a respective Allocation will not be eligible for another placement when the Agent is being drained.  
The `restart` policy is more of an interest because in principle another Agent might be eligible for taking the respective allocation. However, the process of the Nomad agent being restarted in i.e. a deployment is hindering it: The Agent gets the shutdown signal by `systemd`. The agents become ineligible. At the end of the `drain_on_shutdown` `deadline`, the allocations are stopped, the agents restart and the Nomad server tries to place the allocations on another agent (immediately, not after the `reschedule` `delay`). Because at that moment all agents are ineligible and restarting, the Allocation cannot be scheduled and Nomad loses the Job permanently.  
We want to address this circumstance, on the one hand by adjusting the deployment to restart one agent after the other (`codeocean-terraform#3`), and on the other hand by recreating a runner in Poseidon when a job fails due to an unknown reason (#612).

MrSerth: Indeed, thanks for the summary. Based on these thoughts, we agreed to keep the limits currently defined in this PR. Therefore, it can be merged.

PR #633, opened 2024-07-23
Refine updateFileSystem logging messages

### LINKED ISSUE #619: Refactor updateFileSystem Error logging
From the broad issue `Could not perform the requested updateFileSystem.` that captures every error in the file copy process we want to exclude the errors `unexpected EOF`, `node down`, `no allocation found`, and `Unknown allocation`.  
This benefits the proper prioritization of the different causes.

Additionally, the error `node down` should no longer be logged, but instead, Poseidon should respond to CodeOcean with the `ErrorNomadInternalServerError`.

mpass99:
to be able to differentiate between the different errors that occur.

Closes #619

### DIFF
--- a/internal/api/runners.go
+++ b/internal/api/runners.go
@@ -12,7 +12,9 @@ import (
 
 	"github.com/google/uuid"
 	"github.com/gorilla/mux"
+	nomadApi "github.com/hashicorp/nomad/api"
 	"github.com/openHPI/poseidon/internal/config"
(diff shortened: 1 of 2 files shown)


MrSerth (dismissed):
Awesome, thank you! That's great to see and I really expect it will aid us in identifying further issues and prioritize them 💪

MrSerth (approved):

PR #634, opened 2024-07-23
Group Sentry issues

mpass99:
by throwing the log warning at the common level of the Execute call and not at multiple places relying on it.

Related to #589

### DIFF
--- a/internal/api/runners.go
+++ b/internal/api/runners.go
@@ -147,9 +147,6 @@ func (r *RunnerController) updateFileSystem(writer http.ResponseWriter, request 
 	case errors.Is(err, nomadApi.NodeDownErr):
 		entry.Debug("Nomad Node Down while updateFileSystem")
 		writeInternalServerError(request.Context(), writer, err, dto.ErrorNomadInternalServerError)
-	case errors.Is(err, io.ErrUnexpectedEOF):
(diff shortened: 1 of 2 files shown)


MrSerth (dismissed):
Thanks for moving the error handler; I hope this will make it easier to recognize the error by allowing a global count for this type of race condition. 👍

MrSerth:
Once we merged and deployed this issue, we should probably follow-up and sort / merge some of the Sentry issues, too. Can you please do so (since you know better which ones are related)?

MrSerth (approved):

MrSerth:
I just deployed the latest version of Poseidon to production and staging, allowing us to take care of the Sentry issues (merging, ...).

mpass99:
I just resolved [POSEIDON-3W](https://codeocean.sentry.io/issues/4415169628) and [POSEIDON-3V](https://codeocean.sentry.io/issues/4414751740). This allows us to notice when these high-level issues reoccur because of a different low-level error.
I hope to see another `unexpected EOF` error soon creating its own Sentry issue so we can verify that the error check of this PR is working.

MrSerth:
Thanks! As mentioned in the ticket #589, we need to make some further adjustments before resolving [POSEIDON-3W](https://codeocean.sentry.io/issues/4415169628) again (since it was regressed).

PR #635, opened 2024-07-23
Add Exec log statements

mpass99:
to determine whether the `context.Done()` races with the finishing of the Exec call.

Related to #589


mpass99:
Obsolete with the new findings

PR #641, opened 2024-08-07
Fix Group Sentry issue

### LINKED ISSUE #590: Error executing the request
Sentry Issue: [POSEIDON-3W](https://codeocean.sentry.io/issues/4415169628/?referrer=github_integration)

```
Error executing the request
```

- `error executing command in allocation:`**`unexpected EOF`**
  - Verify if a network issue caused this issue or if the runner stopped unexpectedly

mpass99:
Resolves #590 

This time, I checked it's working 🚀

### DIFF
--- a/internal/nomad/api_querier.go
+++ b/internal/nomad/api_querier.go
@@ -138,7 +138,7 @@ func (nc *nomadAPIClient) Execute(ctx context.Context, runnerID string, cmd stri
 	case errors.Is(err, context.Canceled):
 		log.WithContext(ctx).Debug("Execution canceled by context")
 		return 0, nil
-	case errors.Is(err, io.ErrUnexpectedEOF):
+	case errors.Is(err, io.ErrUnexpectedEOF), strings.Contains(err.Error(), io.ErrUnexpectedEOF.Error()):
 		log.WithContext(ctx).WithError(err).Warn("Unexpected EOF for Execute")
 		return 0, nil
 	case strings.Contains(err.Error(), "Unknown allocation"):


MrSerth (approved):
Awesome, thanks for checking! Let's merge and deploy 🚀

MrSerth:
After deploying these changes, we saw a new issue [POSEIDON-63](https://codeocean.sentry.io/issues/5690043317/) as expected. While [POSEIDON-63](https://codeocean.sentry.io/issues/5690043317/) is for `stdout`, the existing issue [POSEIDON-3V](https://codeocean.sentry.io/issues/4414751740/) was reopened for `stderr`. Is this what we expected (and thus are now keeping both issues open)?

mpass99:
No, this is not what we expected 😞 

Sentry is prioritizing the Stack Trace over the Title to group the issues. According to [the documentation](https://docs.sentry.io/product/issues/grouping-and-fingerprints/) we can:

- Introduce a Fingerprinting Rule
  - `message:"Unexpected EOF for Execute*" -> nomad-unexpected-eof`
  - Maybe: `title:"Unexpected EOF for Execute" -> nomad-unexpected-eof` (undocumented)
- SDK-side Fingerprinting

I added the first option for the Sentry Poseidon Project. But, we also can switch to the second option to improve visibility.
We might also add a comment in the Poseidon Error Case Handling referencing the upstream issue.

MrSerth:
Thanks for taking another look here and modifying the fingerprinting options. Just to be clear: What is the expected result? A single issue regardless of `stdout`, `stderr`, file copy, ... for this "Unexpected EOF" issue?

Regarding the two options (fingerprint rule on sentry.io and SDK-side fingerprinting): I don't have a strong opinion for either. The fingerprint rule seems to be easier to implement (and change), but it is also easier to forget. Thus, the SDK-side fingerprinting seems more "permanent", but couples the error reporting more with Sentry. In the end, I would probably prefer the SDK-side fingerprinting in this case, just to ensure we don't have two places where we would need to update.

mpass99:
> A single issue regardless of stdout, stderr, file copy, ... for this "Unexpected EOF" issue?

Yes, thanks for clarifying that. I thing, the described behavior would be most beneficial for tracking the error.

> I would probably prefer the SDK-side fingerprinting

I had a look at [the SDK-side fingerprinting](https://docs.sentry.io/platforms/go/usage/sdk-fingerprinting/) and was a bit disappointed. To me, it seems, there is no option to specify the fingerprint on a per-error basis. Instead, a handler is specified that has to filter all events to identify the right one. This would result in our error being thrown at the API querier, but the fingerprint is adjusted in the main file. This does not improve the visibility as expected. What do you think?

In #643, I fixed the context usage which should result in a different stack trace for Sentry. This might:
- Fix the grouping with Sentry's default mechanism.
- Allow a grouping via [Stack Trace Rules](https://docs.sentry.io/concepts/data-management/event-grouping/stack-trace-rules/)
- Still, allow the Server-side Fingerprinting Rule.

It's a pity that we have to do that many PRs for the same issue due to the uneasy local reproduction due to the involvement of the Nomad and Sentry servers.

Issue #628, opened 2024-07-08
Error loading runner portMappings in Sentry context


MrSerth:
Despite running code successfully, the sentry context is not set correctly [[see this example]](https://codeocean.sentry.io/performance/trace/19666d3dc5ab4091a14f4313571a891c/?node=span-540302fc41015b80&node=txn-b6ceb102e7dd4486b6b88c5147376765&node=txn-9db02d0981b8470cbeb65342cd916540):

```
error loading runner portMappings: error querying allocation for runner 29-c8c06eee-381c-11ef-bc48-fa163e1390db: no allocation found
```

<img width="1374" alt="Bildschirmfoto 2024-07-08 um 10 27 18" src="https://github.com/openHPI/poseidon/assets/7300329/951418cb-2d33-4aa3-820a-556c1536e930">

We should investigate how this could happen and fix the underlying issue.

mpass99:
In the context of #643, we have found the underlying issue of this behavior.

It is caused by our Sentry hook modifying either the global or the request-based scope.

https://github.com/openHPI/poseidon/blob/4ec8aaaf18f2aadf0034599338d7a73a770d14d9/pkg/logging/sentry_hook.go#L25-L33

When checking [recent traces](https://codeocean.sentry.io/performance/trace/3768552ee1054b2f85044dd223a2b1eb), we find that all have the error `synchronize runners failed: nomad Event Stream failed!: error receiving events: unexpected EOF` attached even if the [last occurrence of it](https://codeocean.sentry.io/issues/4584430575/events/7236bb557cfe48ca8436b603ca0d8f70) was four days before.

This is because the context attached to this error has no request-based scope attached, so the global one is used. In this case, the scope was not cleared for multiple days.

We will fix this issue by using [Local Scopes](https://docs.sentry.io/platforms/go/enriching-events/scopes/#using-with-scope)

MrSerth:
Sounds plausible, thank you. Let's merge (and deploy) the changes from #647 and observe the situation.

PR #643, opened 2024-08-09
Refactor Unexpected EOF Sentry handling

mpass99:
by fixing the usage of the wrong context, thus, adding an entry in Sentry's Stack trace. Also, we added a comment referencing the upstream issue regarding the unexpected EOF error.

Related to #641

### DIFF
--- a/internal/nomad/api_querier.go
+++ b/internal/nomad/api_querier.go
@@ -118,7 +118,9 @@ func (nc *nomadAPIClient) Execute(ctx context.Context, runnerID string, cmd stri
 
 	var exitCode int
 	span := sentry.StartSpan(ctx, "nomad.execute.exec")
+	defer span.Finish()
 	span.Description = "Execute Command in Allocation"
+	span.SetData("command", cmd)
 
 	debugWriter := NewSentryDebugWriter(span.Context(), stdout)
 	commands := []string{"/bin/bash", "-c", cmd}
@@ -126,20 +128,20 @@ func (nc *nomadAPIClient) Execute(ctx context.Context, runnerID string, cmd stri
 		Exec(span.Context(), allocation, TaskName, tty, commands, stdin, debugWriter, stderr, nil, nil)
 	debugWriter.Close(exitCode)
 
-	span.SetData("command", cmd)
-	span.Finish()
-
 	switch {
 	case err == nil:
 		return exitCode, nil
 	case websocket.IsCloseError(errors.Unwrap(err), websocket.CloseNormalClosure):
-		log.WithContext(ctx).WithError(err).Info("The exit code could not be received.")
+		log.WithContext(span.Context()).WithError(err).Info("The exit code could not be received.")
 		return 0, nil
 	case errors.Is(err, context.Canceled):
-		log.WithContext(ctx).Debug("Execution canceled by context")
+		log.WithContext(span.Context()).Debug("Execution canceled by context")
 		return 0, nil
 	case errors.Is(err, io.ErrUnexpectedEOF), strings.Contains(err.Error(), io.ErrUnexpectedEOF.Error()):
-		log.WithContext(ctx).WithError(err).Warn("Unexpected EOF for Execute")
+		// The unexpected EOF is a generic Nomad error. However, our investigations have shown that all the current
+		// events of this error are caused by fsouza/go-dockerclient#1076. Because this error happens at the very end,
+		// it does not affect the functionality. Therefore, we don't propagate the error.
+		log.WithContext(span.Context()).WithError(err).Warn("Unexpected EOF for Execute")
 		return 0, nil
 	case strings.Contains(err.Error(), "Unknown allocation"):
 		return 1, ErrNomadUnknownAllocation
--- a/internal/nomad/api_querier.go
+++ b/internal/nomad/api_querier.go
@@ -141,7 +141,8 @@ func (nc *nomadAPIClient) Execute(ctx context.Context, runnerID string, cmd stri
 		// The unexpected EOF is a generic Nomad error. However, our investigations have shown that all the current
 		// events of this error are caused by fsouza/go-dockerclient#1076. Because this error happens at the very end,
 		// it does not affect the functionality. Therefore, we don't propagate the error.
-		log.WithContext(span.Context()).WithError(err).Warn("Unexpected EOF for Execute")
+		log.WithContext(span.Context()).WithError(err).
+			WithField(logging.SentryFingerprintFieldKey, "nomad-unexpected-eof").Warn("Unexpected EOF for Execute")
 		return 0, nil
 	case strings.Contains(err.Error(), "Unknown allocation"):
 		return 1, ErrNomadUnknownAllocation
--- a/pkg/logging/sentry_hook.go
+++ b/pkg/logging/sentry_hook.go
@@ -13,12 +13,23 @@ import (
 // Consider replacing this with a more feature rich, additional dependency: https://github.com/evalphobia/logrus_sentry
 type SentryHook struct{}
 
-const SentryContextKey = "Poseidon Details"
+const (
+	SentryContextKey          = "Poseidon Details"
+	SentryFingerprintFieldKey = "sentry-fingerprint"
+)
 
 var ErrHubInvalid = errors.New("the hub is invalid")
 
 // Fire is triggered on new log entries.
 func (hook *SentryHook) Fire(entry *logrus.Entry) (err error) {
+	hub := handleEntryContext(entry)
+	hub.WithScope(func(scope *sentry.Scope) {
+		err = handleLogEntry(entry, hub, scope)
+	})
+	return err
+}
+
+func handleEntryContext(entry *logrus.Entry) *sentry.Hub {
 	var hub *sentry.Hub
 	if entry.Context != nil {
 		hub = sentry.GetHubFromContext(entry.Context)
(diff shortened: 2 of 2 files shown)


MrSerth:
Thanks for looking into this. To be honest, I am not sure whether this will resolve the issue in the end. 

For one, I was wondering what the difference between `ctx` and `span.Context()` is here (`ctx` is the request context and `span.Context()` the validity of the current span)? Is this why you are also deferring `span.Finish()`?

Finally, I fear that the changes are not enough to get the grouping right.

Why didn't you decide to use `ConfigureScope` with `SetFingerprint` as [documented by Sentry](https://docs.sentry.io/platforms/go/usage/sdk-fingerprinting/#basic-example)?

```golang
sentry.ConfigureScope(func(scope *sentry.Scope) {
	scope.SetFingerprint([]string{"dev-test-unexpected-eof"})
})
```

I tried so, and it worked instantly for execute with both `stderr` and `stdout`. However, the file copy didn't want to apply the same grouping, and I had to come up with the following workaround:

```golang
sentry.GetHubFromContext(ctx).ConfigureScope(func(scope *sentry.Scope) {
	scope.SetFingerprint([]string{"dev-test-unexpected-eof"})
})
log.WithContext(ctx).WithError(err).Warn("Test Grouping for Execute")
```

<details>
<summary>Broader Context for testing</summary>

```golang
# ...

span.SetData("command", cmd)
span.Finish()

switch {
case err == nil:
	sentry.GetHubFromContext(ctx).ConfigureScope(func(scope *sentry.Scope) {
		scope.SetFingerprint([]string{"dev-test-unexpected-eof"})
	})
	log.WithContext(ctx).WithError(err).Warn("Test Grouping for Execute")
	return exitCode, nil
case websocket.IsCloseError(errors.Unwrap(err), websocket.CloseNormalClosure):
	log.WithContext(ctx).WithError(err).Info("The exit code could not be received.")
	return 0, nil
case errors.Is(err, context.Canceled):
	log.WithContext(ctx).Debug("Execution canceled by context")
	return 0, nil
case errors.Is(err, io.ErrUnexpectedEOF), strings.Contains(err.Error(), io.ErrUnexpectedEOF.Error()):
	log.WithContext(ctx).WithError(err).Warn("Unexpected EOF for Execute")
	return 0, nil
case strings.Contains(err.Error(), "Unknown allocation"):
	return 1, ErrNomadUnknownAllocation
default:
	return 1, fmt.Errorf("error executing command in allocation: %w", err)
}
```

</details>

To verify that the fingerprinting is correctly applied, I added the following log statement in `pkg/logging/sentry_hook.go`:

<details><summary>Code Snippet</summary>
<p>


```go
# at the end of `Fire`...

	field := reflect.ValueOf(scope).Elem().FieldByName("fingerprint")
	fmt.Println(GetUnexportedField(field), event)

	hub.CaptureEvent(event)
	return nil
}

func GetUnexportedField(field reflect.Value) interface{} {
	return reflect.NewAt(field.Type(), unsafe.Pointer(field.UnsafeAddr())).Elem().Interface()
}
``` 

</p>
</details> 

Besides the questions raised above, my next question would be why the two routes `/files` and `/execute` behave differently regarding their Sentry "hub", "context", ... (i.e., the first requires `sentry.GetHubFromContext(ctx).ConfigureScope(...)`, while the second works with `sentry.ConfigureScope(...)`).

As soon as the custom fingerprinting (with `GetHubFromContext `) is enabled, the following event is created: [POSEIDON-6E](https://codeocean.sentry.io/issues/5694734855/). As I tested through CodeOcean, the number of events is expected (one for the file copy + 2 test files x 2 streams [stdout, stderr] = 1 + 2 x 2 = 5).

It shows the events for different routes together ...
Bildschirmfoto 2024-08-09 um 16 17 25

... and acknowledges the custom fingerprinting (at the end of the "Details" tab)
Bildschirmfoto 2024-08-09 um 16 17 41

mpass99:
Thank you for diving into this thus allowing me to gain further insights!

> ctx is the request context and span.Context() the validity of the current span?

Yeah, `ctx` is the request context. `span.Context()` is also the request context but with the span stored inside.  
You are right, this doesn't change anything. Here, I got bamboozled by my own Fingerprinting rule that was still active.

> Why didn't you decide to use `ConfigureScope` with `SetFingerprint` as [documented by Sentry](https://docs.sentry.io/platforms/go/usage/sdk-fingerprinting/#basic-example)?

I decided against it because `ConfigureScope` defines the global scope of the Sentry SDK. Therefore, all events would be grouped into one single event. Without the `{{ default }}` option mentioned in the documentation, this method seems to have little value.

The difference to `sentry.GetHubFromContext(ctx).ConfigureScope(...)` is that the scope of the context is being used. This improves the situation, but still, the scope is configured on a per-request basis and not on a per-error basis. This means that the [StdOut](https://codeocean.sentry.io/issues/5696154057/events/016b7c3331c64e7c9d3d6300f2681da5) and [StdErr](https://codeocean.sentry.io/issues/5696154057/events/62cbbc9c98d543da9900d0ada13a6b08) Error is grouped together, but also [a totally unrelated error in the same request](https://codeocean.sentry.io/issues/5696154057/events/e40e51407cb747e38673639601750e84).

> Why [do] the two routes `/files` and `/execute` behave differently regarding their Sentry "hub", "context", ... (i.e., the first requires `sentry.GetHubFromContext(ctx).ConfigureScope(...)`, while the second works with `sentry.ConfigureScope(...)`).

I assume that the second also does not directly work with `sentry.ConfigureScope(...)`. This is because this function modifies the global scope, but our custom Sentry hook prefers the context scope over the global scope. Only on the following request, the context scope is cloned from the global scope.

I assume that the wrong usage in our Sentry Hook is also the underlying error of #628.

> Add a custom WithFingerprint method to the logrus chain

That's a nice idea, but with Golang [we can only](https://stackoverflow.com/questions/41109708/cant-define-receiver-from-another-package-in-go) define a method on a type defined in the same package, not on the `*logrus.Entry` from the `logrus` package.

However, we can add a field that is automatically mapped to the fingerprint. I would implement this together with the refactoring of our Sentry Hook (e.g. clone the scope), once we conclude the investigation.

<details><summary>Active Rules found in the Project Configuration</summary>
<p>

Do I remember correctly that none of these was set before? Have you added the Stack Rules or did I a long time ago and forget about it due to the missing visibility?

```
Fingerprint Rules:
message:"Unexpected EOF for Execute*" -> nomad-unexpected-eof


Stack Rules:
stack.abs_path:**/go/pkg/mod/github.com/sirupsen/**      ^-app -app ^-group -group
stack.abs_path:**/sirupsen/logrus*/**/*.go               ^-app -app ^-group -group
stack.function:(*Entry).Warn                             ^-app -app ^-group -group
stack.abs_path:**/pkg/logging/sentry_hook.go             ^-app -app ^-group -group
```

</p>
</details>

MrSerth:
Thanks for taking a moment to reconsider and replying -- that was very helpful for me, too!

> Here, I got bamboozled by my own Fingerprinting rule that was still active.

I forgot the fingerprinting rule, too, when experimenting with this feature (and only later changed the error message)...

> I decided against it because `ConfigureScope` defines the global scope of the Sentry SDK. 

I wasn't aware that this changes the global config (and of course, this is not what we want).

> The difference to `sentry.GetHubFromContext(ctx).ConfigureScope(...)` is that the scope of the context is being used. This improves the situation, but still, the scope is configured on a per-request basis and not on a per-error basis. This means that the [StdOut](https://codeocean.sentry.io/issues/5696154057/events/016b7c3331c64e7c9d3d6300f2681da5) and [StdErr](https://codeocean.sentry.io/issues/5696154057/events/62cbbc9c98d543da9900d0ada13a6b08) Error is grouped together, but also [a totally unrelated error in the same request](https://codeocean.sentry.io/issues/5696154057/events/e40e51407cb747e38673639601750e84).

Okay, better 😅, but still not ideal. Then, what about either:

- reverting the fingerprint back to `{{ default }}` after sending the event
- or using a [temporary scope](https://docs.sentry.io/platforms/go/enriching-events/scopes/#using-with-scope) (sounds better, IMHO).

> I assume that the second also does not directly work with `sentry.ConfigureScope(...)`. 

Yes, you are right. With `sentry.ConfigureScope(...)`, only the second request *after starting Poseidon* (regardless of the specific request) is getting fingerprinted. If another configuration of the scope occurred previously, the `/files` route is immediately fingerprinted "correctly". If, however, files are not copied over and tests are run immediately (`/execute`), the first request (two calls, one for `stdout` and one for `stderr`) are not fingerprinted. Only the second one is. 

Since I was "debugging" yesterday, I was restarting Poseidon frequently, Thus, my first request always was the `/files` request which only occurred once...

> I assume that the wrong usage in our Sentry Hook is also the underlying error of https://github.com/openHPI/poseidon/issues/628.

Could be 🤷 

> That's a nice idea, but with Golang [we can only](https://stackoverflow.com/questions/41109708/cant-define-receiver-from-another-package-in-go) define a method on a type defined in the same package, not on the `*logrus.Entry` from the `logrus` package.

Alright 😁

> However, we can add a field that is automatically mapped to the fingerprint. I would implement this together with the refactoring of our Sentry Hook (e.g. clone the scope), once we conclude the investigation.

Sounds good, I am a fan to fix the underlying issue :+1:

> Do I remember correctly that none of these was set before? Have you added the Stack Rules or did I a long time ago and forget about it due to the missing visibility?

No, you don't remember that correctly 😉. You added these four stack rules yourself, as part of https://github.com/openHPI/poseidon/pull/412#issuecomment-1647885591. I think, they are still needed, aren't they?

mpass99:
> I think, they are still needed, aren't they?

I agree, the Stack Rules are still needed. However, when we have stored rules there anyway, our [initial argument](https://github.com/openHPI/poseidon/pull/641) that "we don't have two places where we would need to update" is not fixed by an SDK-side fingerprint solution.

- I'm fine with doing the Fingerprinting anyway SDK-side and adding documentation for visibility of the stack rules.
- Or we do it in the Sentry Backend and add documentation for visibility fingerprinting rules and the stack rules.

MrSerth:
> When we have stored rules there anyway, our [initial argument](https://github.com/openHPI/poseidon/pull/641) that "we don't have two places where we would need to update" is not fixed by an SDK-side fingerprint solution.

That's (for me) only partially true, but in general I agree. Having the Stack Rules online (and not in the code) is not ideal either. However, to me, it feels like this is more "static" given the general setup of our Sentry  integration. We also need to configure a few other things there.

Contrary, the fingerprinting feels more "dynamic" to me, since we are more frequently adding, modifying, and removing error strings. That having said, I would still go with the SDK-side fingerprinting if you find an easy solution for that.

MrSerth:
Thank you for these changes, this looks pretty good and clean now. I just have one final question: Why did you decide to continue using `span.Context()` rather than `ctx` for `og.WithContext`? Is this to connect the error to the inner most span available? Just asking again, since your previous reply looked like it is pretty interchangeable.

> Yeah, `ctx` is the request context. `span.Context()` is also the request context but with the span stored inside.
> You are right, this doesn't change anything. Here, I got bamboozled by my own Fingerprinting rule that was still active.

mpass99:
Hmm, I still think it is interchangeable. However, from a conceptual point of view, using the most inner context makes sense.
In the future, Sentry might use the trace information it stores in the context, to connect traces and errors more easily.

However, now, I don't see an indicator for Sentry doing so. Also, we don't use the context in our `StartSpan` helper function:
https://github.com/openHPI/poseidon/blob/06deab08b1f10a8db30a896d846ab7f2d2b3924a/pkg/logging/sentry_hook.go#L69-L74

Although, it seems quite odd to pass the context to the callback if the callback could just use the context passed to `StartSpan`. For me, it looks like, `StartSpan` was meant to pass `span.Context()` to the callback.

However, due to this inconsistency, I'm fine with just using `ctx` rather than `span.Context()` for `log.WithContext` (dropping fc7312d5b60ec77018a5bdbcfe92203dca748d0f).

MrSerth:
Thanks for detailing your thoughts on this. Let's keep `span.Context()` here, to use it "correctly" from a conceptional point of view (regardless of the actual context passed by `StartSpan`).

> Although, it seems quite odd to pass the context to the callback if the callback could just use the context passed to StartSpan. For me, it looks like, StartSpan was meant to pass span.Context() to the callback.

Feels so, yes. Shall we try that with a follow-up PR?

mpass99:
> Feels so, yes. Shall we try that with a follow-up PR?

Sounds good. I opened #654

To gain some insights, I had a quick glimpse into the Senty SDK. It actually connects the error to a span.
[Here](https://github.com/getsentry/sentry-go/blob/11d3c795ddc3d0a28ddb26755584ba4a7d314e7e/tracing.go#L197-L199), it sets the trace information in the scope. [Here](https://github.com/getsentry/sentry-go/blob/11d3c795ddc3d0a28ddb26755584ba4a7d314e7e/scope.go#L397), the trace information of the span scope is applied to the event. This already works because `ApplyToEvent` is called in `CaptureEvent` and [we call `CaptureEvent`](https://github.com/openHPI/poseidon/blob/7efee3d440f7e2282753a65d3da6563f1b074d42/pkg/logging/sentry_hook.go#L24) with the same hub/scope the tracing to store the trace information.

However, as we identified before, the used scope is request-based. Therefore, I really wonder how Sentry deals with multiple goroutines. For example, for each user's WebSocket request, we have two Nomad connections, each using its own spans.
https://github.com/openHPI/poseidon/blob/7efee3d440f7e2282753a65d3da6563f1b074d42/internal/nomad/api_querier.go#L120
With this, the two Goroutines/Execute-calls overwrite their trace information stored in the request-based scope prohibiting a reliant error matching.

Sentry might fix this by separating the scopes e.g. when starting a new span, [like it already does for each HTTP request](https://github.com/getsentry/sentry-go/blob/11d3c795ddc3d0a28ddb26755584ba4a7d314e7e/http/sentryhttp.go#L89-L90):

```
hub = sentry.CurrentHub().Clone()
ctx = sentry.SetHubOnContext(ctx, hub)
```

MrSerth (approved):
Thanks for investigating and looking into the Sentry SDK. I think, for some use cases it might be fine to have the same scope in a Goroutine, but probably not in those complex scenarios with further WebSocket requests.

If I understood this right, the question you raised in #654 about the concurrency issue is directly related to this finding with Goroutines. Hence, I think this PR is fine and complete, allowing us to merge it (and continue the last aspect in the linked PR #654).

PR #646, opened 2024-08-13
Update Dependencies

MrSerth:
Update dependencies and Nomad in CI


go.mod
```diff
@@ -1,15 +1,15 @@
 module github.com/openHPI/poseidon
 
-go 1.22.4
+go 1.22.6
```
mpass99, line 3: Are we ready for [Go 1.23](https://go.dev/doc/go1.23)? 🚀

mpass99 (approved):

go.mod
MrSerth: Sure, let's upgrade in a follow-up PR (shipping small increments 🤓).

Since upgrading Golang to 1.23 might also require a check (and change?) on our deployment machines, I'll create a new ticket.

PR #647, opened 2024-08-14
Fix Sentry Hook Global Scope

### LINKED ISSUE #628: Error loading runner portMappings in Sentry context

mpass99:
that led to errors being duplicated in the following events sharing the same scope.

Now, every event has its own local scope.

Closes #628

### DIFF
--- a/pkg/logging/context_hook.go
+++ b/pkg/logging/context_hook.go
@@ -6,7 +6,7 @@ import (
 )
 
 // ContextHook logs the values referenced by the of dto.LoggedContextKeys.
-// By default Logrus does not log the values stored in the passed context.
+// By default, Logrus does not log the values stored in the passed context.
 type ContextHook struct{}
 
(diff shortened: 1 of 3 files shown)


MrSerth (approved):
Thanks for fixing the underlying issue and adding a test 😻

PR #652, opened 2024-08-16
Fix Content Length Writer

### LINKED ISSUE #595: Could not write JSON response
Sentry Issue: [POSEIDON-5M](https://codeocean.sentry.io/issues/5210781506/?referrer=github_integration)

```
Could not write JSON response: http: wrote more than the declared Content-Length
```

This happened when we as administrators opened some data below `/proc`.

mpass99:
- The `Content-Length`-Header did not get removed in the error handling
- The Content-Length Writer fired the HTTP Headers unnecessarily early impeding sending of an error message.

Closes #595


MrSerth (approved):
Thanks, looks good and fixes the actual problem (I reproduced it first with CodeOcean locally and then upgraded Poseidon to see that it got resolved) :+1:

PR #655, opened 2024-08-17
Ungroup Sentry Error

mpass99:
that contains both Environment Recovery Errors and Runner Recovery Errors.

Related to #650


MrSerth (dismissed):
Looks good, let's give it a test.

Issue #651, opened 2024-08-16
Sentry: the runner will be destroyed and replaced


MrSerth:
[POSEIDON-3W](https://codeocean.sentry.io/issues/4415169628/) was regressed due to the following error: „the runner will be destroyed and replaced: the destruction should not cause external changes“.

<hr>

@mpass99 wrote on Aug 16th, 15:01 CEST:

I'm glad that we have gained visibility on such rare errors. The last event happened because of an unattended upgrade that restarted some Nomad servers. In the process of reestablishing the Event Stream and recovering all runners, all runners got recreated. In the case of this execution, the recreation happened during a running execution connection.
This error is nothing we would not expect. We can either

- track (and ignore) it in this issue
- or catch it in Poseidon
  - and throw a different error
  - and throw no error but lose the visibility on the number of executions that get aborted this way.

MrSerth:
Thanks for analyzing and identifying the root cause for the behavior we observed. 

I have two thoughts on this (or similar) issue(s):

1. First, it would be great if we could quickly identify Sentry issues we only have to "track" some behavior (i.e., count) and where no immediate action is required. Can we manually append some label or tag to these?
2. While I understand the argumentation behind ignoring the issue (either with a Sentry event or without), this doesn't feel right to me. From a user perspective, the execution requested failed and a red error was shown. Then, it shouldn't make a difference to a user whether this is a "planned" event (such as a process restart) or some other failure. Therefore, I would tend to dig a little further here.

The underlying issue, if I got this right, is that the runner got recreated after a Nomad agent restart. Then, I am wondering about the effectiveness of our `drain_on_shutdown` strategy on Nomad and whether Poseidon could track these events, too. Since a live migration of Docker containers (with their changes on disk and in RAM) is not yet supported by Nomad, we might need to "exclude" some runners for new executions as soon as the respective Nomad agent is about to shut down. Is there some indication this could work and would solve the underlying issue?

mpass99:
> Can we manually append some label or tag to these?

We can add a Sentry Tag or adjust the Sentry Log Level (info/debug). The tag can be more self-explanatory, but the Log Level would be visible in the Issue overview.

> the runner got recreated after a Nomad agent restart

In this case, just the Nomad Server was restarted. This led to [the Event Stream being aborted](https://codeocean.sentry.io/issues/4584430575/events/3a603099fce045548f8f650e14fe691b). When re-establishing the event stream, we do an environment and runner recovery:
- Load Environments
[...]

PR #661, opened 2024-08-20
Separate Sentry Issues

mpass99:
to allow for the differentiation of local recovery errors from generic execution failures.

Related to #651 

- We might want to decrease the Sentry Log Level for this issue.

### DIFF
--- a/internal/api/ws/codeocean_writer.go
+++ b/internal/api/ws/codeocean_writer.go
@@ -114,6 +114,10 @@ func (cw *codeOceanOutputWriter) Close(info *runner.ExitInfo) {
 		message := "the allocation stopped as expected"
 		log.WithContext(cw.ctx).WithError(info.Err).Trace(message)
 		cw.send(&dto.WebSocketMessage{Type: dto.WebSocketOutputError, Data: message})
+	case errors.Is(info.Err, runner.ErrDestroyedAndReplaced):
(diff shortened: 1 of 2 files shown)


internal/api/ws/codeocean_writer.go
```diff
@@ -114,6 +114,10 @@ func (cw *codeOceanOutputWriter) Close(info *runner.ExitInfo) {
 		message := "the allocation stopped as expected"
 		log.WithContext(cw.ctx).WithError(info.Err).Trace(message)
 		cw.send(&dto.WebSocketMessage{Type: dto.WebSocketOutputError, Data: message})
+	case errors.Is(info.Err, runner.ErrDestroyedAndReplaced):
+		errorMessage := "Runner recovery stopped request execution"
```
MrSerth, line 118: request or requested? The one is related to the HTTP term (and technically correct), but I find the other one easier to read. 

```suggestion
		errorMessage := "Runner recovery stopped requested execution"
```

MrSerth (commented):
Looks good, and (without testing) I assume it separates the issue as desired. I ams till not sure whether "that's it" for the linked issue #651, but it's helpful anyway, I'd say.

mpass99:
With #667, we stop replacing the runner reference on recovery. Therefore, the error that is handled in this PR should no longer occur.

Issue #623, opened 2024-06-27
Fix flaky e2e test


MrSerth:
- Failed: https://github.com/openHPI/poseidon/actions/runs/9698681202/attempts/1
- Successful: https://github.com/openHPI/poseidon/actions/runs/9698681202/attempts/2

```
websocket_test.go:363: 
    	Error Trace:	/home/runner/work/poseidon/poseidon/tests/e2e/websocket_test.go:363
    	            				/home/runner/work/poseidon/poseidon/tests/e2e/websocket_test.go:170
    	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:115
    	Error:      	Not equal: 
    	            	expected: 1
    	            	actual  : 0
    	Test:       	TestE2ETestSuite/TestCommandMakeEnvironmentVariables/0
```

- Failed: https://github.com/openHPI/poseidon/actions/runs/10672601360/job/29580481134
- Successful: https://github.com/openHPI/poseidon/actions/runs/10672601360/job/29580604118

<details><summary>Unexpected EOF</summary>
<p>

```
=== RUN   TestE2ETestSuite/TestGetFileContent_Nomad/Permission_Denied
{"environment_id":"0","error":"unexpected EOF","level":"warning","msg":"Unexpected EOF for Execute","package":"nomad","runner_id":"0-8a48a4a7-6969-11ef-a609-000d3a600657","sentry-fingerprint":["nomad-unexpected-eof"],"time":"2024-09-02T20:25:50.041434Z"}
    runners_test.go:473: 
        	Error Trace:	/home/runner/work/poseidon/poseidon/tests/e2e/runners_test.go:473
        	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:115
        	Error:      	Not equal: 
        	            	expected: 424
        	            	actual  : 200
        	Test:       	TestE2ETestSuite/TestGetFileContent_Nomad/Permission_Denied
    runners_test.go:476: 
        	Error Trace:	/home/runner/work/poseidon/poseidon/tests/e2e/runners_test.go:476
        	            				/home/runner/go/pkg/mod/github.com/stretchr/testify@v1.9.0/suite/suite.go:115
        	Error:      	"" does not contain "file not found or insufficient permissions"
        	Test:       	TestE2ETestSuite/TestGetFileContent_Nomad/Permission_Denied
```

</p>
</details>

mpass99:
[...]

PR #662, opened 2024-08-20
Fix missing link target

### LINKED ISSUE #596: Could not split link into name and target
Sentry Issue: [POSEIDON-5K](https://codeocean.sentry.io/issues/5210781378/?referrer=github_integration)

```
could not split link into name and target
```

This happened when we as administrators opened some data below `/proc`.

mpass99:
for listing the file system with a link, the executing user does not have enough privileges to read the target.

Closes #596

### DIFF
--- a/pkg/nullio/ls2json.go
+++ b/pkg/nullio/ls2json.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"regexp"
@@ -20,6 +21,7 @@ var (
 	pathLineRegex   = regexp.MustCompile(`(.*):$`)
 	headerLineRegex = regexp.
 			MustCompile(`([-aAbcCdDlMnpPsw?])([-rwxXsStT]{9})(\+?) +\d+ +(.+?) +(.+?) +(\d+) +(\d+) +(.*)$`)
+	ErrLinkTargetPermissionDenied = errors.New("permission denied for reading link target")
 )
 
 const (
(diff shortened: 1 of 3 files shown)


pkg/nullio/ls2json.go
```diff
@@ -163,11 +163,14 @@ func (w *Ls2JsonWriter) parseFileHeader(matches [][]byte) ([]byte, error) {
...
 		const NumberOfPartsInALink = 2
-		if len(parts) == NumberOfPartsInALink {
+		switch len(parts) {
+		case NumberOfPartsInALink:
 			name = dto.FilePath(parts[0])
 			linkTarget = dto.FilePath(parts[1])
-		} else {
-			log.WithContext(w.sentrySpan.Context()).Error("could not split link into name and target")
+		case 1:
+			// This case happens when a user tries to read the target of a link without permission. See #596.
```
MrSerth, line 171: We could improve the comment to notice that we don't do any action for such an entry (no op).

MrSerth (commented):
Thanks, looks good! I tested the branch locally and can confirm that your patch resolves the issue for me.

There is one question left, though: When listing files where this (former) bug triggers, I still have an exit code 1 and thus the following warning: ` level=warning msg="Ignoring error of listing the file system" error="<nil>" exitCode=1 package=runner`. Do we just continue to ignore it (and log as warning) or do we want to do something special (since we know why the error was generated in the first place....)

MrSerth (approved):
Perfect, works like a charm. Thanks!

MrSerth:
I've created an issue for the flaky unit test #669.

MrSerth:
I've also amended ticket #623 for the failed e2e test. It seems like we need to investigate in the test suite, once again.

PR #667, opened 2024-09-02
Fix runner replacement on recovery

### LINKED ISSUE #651: Sentry: the runner will be destroyed and replaced

mpass99:
by now just updating the port mapping.

Closes #651

### DIFF
--- a/internal/runner/nomad_manager.go
+++ b/internal/runner/nomad_manager.go
@@ -240,28 +240,48 @@ func (m *NomadRunnerManager) loadSingleJob(ctx context.Context, job *nomadApi.Jo
 	return newJob, isUsed, nil
 }
 
-// updateUsedRunners handles the cleanup process of updating the used runner storage.
-// This includes the clean deletion of the local references to the (replaced/deleted) runners.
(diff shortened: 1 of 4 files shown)


MrSerth (commented):
Looks good, thanks!

There are still many annotations, mostly from the linter -- please double check these. After fixing all other linter issues, the pipeline should generally pass 😉

MrSerth:
I just observed another occurrence of #669.

MrSerth (approved):
Thanks, looks good! Let's give it a try 😉

PR #671, opened 2024-09-03
Remove deprecated go get -d flag

mpass99:
that now represents the default behavior [[1]](https://pkg.go.dev/cmd/go/internal/get).

The deprecation warning shows up when running `make test`.


MrSerth (approved):
Thanks for updating!

Issue #673, opened 2024-09-04
Nomad Agent Simultaneous Restart Behavior


mpass99:
In #612 we noticed that on a simultaneous restart of all Nomad agents, some Jobs are completely removed and disappear, others are dead but still listed, and some are being restarted.

- Create Reproduction Steps for this behavior
- Create an upstream Issue
  - Why are most of the Jobs not restarted after the agents are ready again?
  - Why do some Jobs are completely removed while others are still listed as `complete`/dead
  - Why do we not receive `Job`-`JobDeregistered` events?

mpass99:
> Create Reproduction Steps for this behavior

<details><summary>Minimal Job Configuration</summary>
<p>

``` json
{
  "ID": "1",
  "Name": "1",
  "Type": "batch",
  "TaskGroups": [
    {
      "Name": "default-group",
      "Count": 1,
      "RestartPolicy": {
        "Attempts": 3,
        "Interval": 3600000000000,
        "Delay": 15000000000,
        "Mode": "fail",
        "RenderTemplates": false
      },
      "ReschedulePolicy": {
        "Attempts": 3,
        "Interval": 21600000000000,
        "Delay": 60000000000,
        "DelayFunction": "exponential",
        "MaxDelay": 240000000000,
        "Unlimited": false
      },
      "Tasks": [
        {
          "Name": "default-task",
          "Driver": "docker",
          "Config": {
            "command": "sleep",
            "force_pull": true,
            "image": "openhpi/co_execenv_python:3.8",
            "network_mode": "none",
            "args": [
              "infinity"
            ]
          }
        }
      ]
    }
  ]
}
```

</p>
</details>

We conducted 5 repetitions with each having 5 jobs running while restarting the agents.  
Interestingly, the recreation does neither count as `Restart` nor as `Rescheduling`, although the events state that the Allocation is being migrated.

| Repetition | Running Jobs after Restart |
|--------|--------|
| [1](https://github.com/user-attachments/files/16886988/nomadEventDump-RestartTogether-Minimal-1-1725524911.txt) | 5/5 |
| [2](https://github.com/user-attachments/files/16886989/nomadEventDump-RestartTogether-Minimal-2-1725525328.txt) | 4/5 |
| [3](https://github.com/user-attachments/files/16886990/nomadEventDump-RestartTogether-Minimal-3-1725525948.txt) | 5/5 |
[...]

PR #680, opened 2024-09-05
Fix prohibiting Nomad Job recreation with delay in between

mpass99:
by not purging a job once its allocation stops (and maybe would be replaced).

Related to #673 

Scenario: Deleting and Creating the environment before each restart
Before PR
1 - 0/5
2 - 0/5
3 - 0/5
4 - 0/5
5 - 0/5

After PR
1 - 5/5
2 - 5/5
3 - 5/5 (one slow recreation)
4 - 5/5
5 - 5/5


MrSerth (approved):
Looks good, thanks! 🪴

PR #681, opened 2024-09-05
Fix OOM Killing Nomad Agent

### LINKED ISSUE #676: Nomad Agents DoS on Migration
In #612, we noticed that the Nomad Agents are being caught in a crash loop, when too many allocations are being migrated (in response to a restart).

We find OOM killing errors (telegraf, systemd) and Nomad errors.

- Investigate if this error is reproducable without telegraf.

mpass99:
by increasing the memory reservation per runner.

Closes #676 

This may not be the most stable solution as different environments (images) might require different amounts of memory. However, with increasing the memory reservation per runner only as many runners are allocated on the agent as it could handle right now (with the given server resources, and the allocation definition).

### DIFF
--- a/internal/environment/template-environment-job.hcl
+++ b/internal/environment/template-environment-job.hcl
@@ -51,7 +51,7 @@ job "template-0" {
 
       resources {
         cpu    = 40
-        memory = 30
+        memory = 50
       }
     }
   }


mpass99:
That's true, but so we might still have the OOM Killing behavior and also still missing visibility on it, preventing detection of such Nomad agent overloadings.

PR #686, opened 2024-09-11
Fix Golangci Lint v1.61

mpass99:
[Some reference](https://dev.to/ccoveille/about-the-gosec-g115-drama-or-how-i-faced-back-integer-conversion-overflow-in-go-1302)


MrSerth (approved):
Thanks for the article and for fixing; that was a good read (for me 🤗).

PR #689, opened 2024-09-13
Catch unsupported Systemd Watchdog

### LINKED ISSUE #644: Add option to configure Systemd Watchdog
Not all operating systems use systemd, and thus don't necessarily support the systemd watchdog. One example would be macOS 💻. However, with the current setup, an *error* is issues that the systemd watchdog is unsupported on each start of Poseidon.

When error reporting to Sentry is enabled, consequentially, each start of Poseidon on macOS reports a new event (which I have to delete manually). It would be great if we could change the current behavior.

I see the following potential approaches:

1. Add a config option to disable the systemd watchdog
2. Lower the level of the "Systemd Watchdog not supported" error (maybe similar to "Systemd Readiness Notification not supported")
3. Couple systemd readiness notifications and systemd Watchdog: Return from `notifySystemd`, when systemd readiness notifications are not supported
4. (Disable the watchdog for non-Linux systems)

mpass99:
Closes #644

### DIFF
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -249,8 +249,11 @@ func notifySystemd(router *mux.Router) {
 	}
 
 	interval, err := daemon.SdWatchdogEnabled(false)
-	if err != nil || interval == 0 {
-		log.WithError(err).Error("Systemd Watchdog not supported")
+	if err != nil {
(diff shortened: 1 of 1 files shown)


MrSerth (approved):
Works like a charm, thank you! 🌈

PR #704, opened 2024-09-27
Fix provideRunner with no managers enabled.

mpass99:
If only AWS was enabled, a 500 error instead of a 404 error was returned. If both Nomad and AWS was disabled, the requests returned with a 200 response but an empty body. A silent nil-dereference was triggered.

Closes https://github.com/orgs/openHPI/projects/4/views/1?pane=issue&itemId=81052193

### DIFF
--- a/cmd/poseidon/main.go
+++ b/cmd/poseidon/main.go
@@ -341,6 +341,13 @@ func createManagerHandler(ctx context.Context, handler managerCreator, enabled b
 	return runnerManager, environmentManager
 }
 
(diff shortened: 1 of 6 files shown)


MrSerth (approved):
Nice, thank you! I tested the functionality and was able to confirm that it now works as expected.

As part of my review process, I also observed that synchronizing any environment to Poseidon was successful when no runner manager was enabled. Hence, I amended your PR to change this behavior and further fix a superfluous header write (due to a missing return statement). With that, I am happy to merge this PR 👍.

